PR #9, opened 2024-10-07
Define an interface for some of the major components

russelldj:
This PR defines an interface for some of the major components. There are a few comments about potential implementation suggestions but not many. I think this functionality would get us a large part of the way toward the MVP we've discussed internally, but some stuff will need to be added to fill in the gaps.


tree_detection_framework/detection/detector.py
```diff
@@ -0,0 +1,74 @@
...
+    def setup(self):
+        # This method should implement setup tasks that are common to all LightningDetectors.
+        # Method-specific tasks should be defered to setup_model
+        raise NotImplementedError()
+
+    @abstractmethod
+    def setup_model(self):
+        """Set up the lightning model, including loading pretrained weights if required"""
+        # Should be implemented in each derived class since it's algorithm-specific
+        raise NotImplementedError()
```
amrithasp02, line 48: I think it's nice that you decided to separate the model-specific setup from the common setup steps with 2 different functions.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -0,0 +1,95 @@
...
+            controlled by the corresponding element in split_fractions.
+
+    """
+    raise NotImplementedError()
+
+
+def create_dataloader(
+    raster_folder_path: PATH_TYPE,
+    chip_size: float,
+    chip_stride: float,
```
amrithasp02, line 36: Right now in sandbox/chip_ortho.py, there is also an option to provide `overlap_percent`. Are we planning to remove this option or handle it when `create_dataloader()` is actually being called since the `overlap` anyway gets converted to a `stride` value before inputting to `GridGeoSampler`.

amrithasp02 (approved):
I think all of these interfaces look great!

tree_detection_framework/preprocessing/preprocessing.py
russelldj: I'm open to discussion on this point but my initial thought was this function should be streamlined as possible. I still think we may want to provide the option to do a `percent_overlap` in the entrypoint function that puts it all together. To me this feels like a good place for this type of "user interface/business logic" sort of stuff.

tree_detection_framework/detection/detector.py
russelldj: I think it's possible that we can remove `setup` in favor of just `__init__` but for the time being I think we can try keeping them separate and see if that makes it any cleaner.

tree_detection_framework/preprocessing/preprocessing.py
amrithasp02: Totally agree

tree_detection_framework/detection/region_detections.py
```diff
@@ -0,0 +1,95 @@
...
+
+        Args:
+            detections (gpd.GeoDataFrame):
+                The dataframe of detected shapes, potentially with additional attributes such as
+                class or confidence. If no CRS is set, the data will be assumed to be in pixel
+                coordinates.
+            pixel_to_CRS_transform (Optional[np.array], optional):
+                Only meaningful if `CRS` is set as well. A 2x3 transform matrix mapping the
+                coordinates in pixels to the coordinates of the CRS. If un-set, will assumed to be
+                the identity transform. Defaults to None.
```
youngdjn, line 29: Alternatively, what if this stores the geospatial coords (in the specified CRS) of the top-left pixel. Then combined with the CRS we'd have the minimal information needed to translate pixels to geospatial and back.

tree_detection_framework/postprocessing/postprocessing.py
youngdjn, line 1: I'd like to include alternatives/tweaks to NMS, like downweighting detections closer to the edges of chips, and optionally simply dropping all detections with centroids in the (overlapping) buffer regions. Would that fit here, and if so, should we specify the interface now or wait?

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -0,0 +1,95 @@
...
+    """Show samples from the dataloader
+
+    Args:
+        dataloader (DataLoader): The dataloader to visualize
+        n_tiles (int): The number of randomly-sampled tiles to show
+    """
+    raise NotImplementedError()
+
+
+def save_dataloader_contents(dataloader: DataLoader, save_folder: PATH_TYPE):
```
youngdjn, line 88: I love the idea of including this! Would there ever be a case where we would want to create a dataloader by loading whatever was saved by this method?

youngdjn: Should there be an option to specify a certain (non-)(random) subset of the dataloader to be saved, for custom external visualization tasks?

youngdjn (approved):

tree_detection_framework/preprocessing/preprocessing.py
russelldj: Yes, sounds good. Maybe include `n_tiles` and `random_sample` arguments?

tree_detection_framework/detection/detector.py
```diff
@@ -0,0 +1,74 @@
...
+        Returns:
+            RegionDetectionsSet: One prediction per tile
+        """
+        # This should not be implemented here unless there are prediction tasks shared across every detector
+        raise NotImplementedError()
+
+
+class LightningDetector(Detector):
+    model: lightning.LightningModule
+    trainer: lightning.Trainer
```
youngdjn, line 37: Remove `trainer` attribute and have the lightning Trainer be defined by the `train` method.

tree_detection_framework/postprocessing/postprocessing.py
russelldj: Yes, this would be another function with similar arguments and returns.

tree_detection_framework/preprocessing/preprocessing.py
russelldj: After the discussion with Derek, I'm going back to the original way you were thinking Amritha. His logic was a user might want to call this function directly in a notebook and we shouldn't make them write the logic there.

tree_detection_framework/detection/detector.py
russelldj: That's really smart, I like that. We could also have a stand-alone function or static method. I guess we really need to figure out how we want to specify our configuration parameters and specifically whether they are all specified when the object is created.

russelldj: One reason to lean in favor of a method or static method would be if we want to parse a library-specific config file. Like if DeepForest provides a training config and we need to map this to a standardized format.

russelldj:
Merged for now, we can always update this interface as we make more progress.

PR #16, opened 2024-10-14
Refactored the chipping script from sandbox into the main module

amrithasp02:
In `tree_detection_framework/preprocessing`:

- I created `derived_geodatasets.py` to hold the CustomRaster and CustomVector datasets. Also moved the plot function inside `CustomRasterDataset`.
- In `preprocessing.py` I implemented the `create_dataloader`, `visualize_dataloader`, and `save_dataloader_contents` functions.

In `tree_detection_framework/entrypoints`:
- Added the script `tile_data.py` to test and perform the preprocessing from command line.


amrithasp02:
I also placed `get_projected_CRS()` in `utils/geospatial.py`. Let me know if you feel the `geospatial.py` needs to be renamed to something more generic.

tree_detection_framework/entrypoints/tile_data.py
```diff
@@ -1 +1,201 @@
...
+    visualize_n_tiles: Optional[int] = None,
+    save_folder: Optional[PATH_TYPE] = None,
+    save_n_tiles: Optional[int] = None,
+    random_sample: bool = False,
+):
+    """
+    Entrypoint script to test preprocessing functions.
+
+    Args:
+        raster_folder_path (PATH_TYPE): Path to the folder or raster files.
```
russelldj, line 34: style nitpick: could be indented to match the other arguments even though it is a short line.

tree_detection_framework/entrypoints/tile_data.py
```diff
@@ -1 +1,201 @@
...
+        save_folder (Optional[PATH_TYPE], optional):
+            Folder to save data to. Will be created if it doesn't exist.
+        save_n_tiles (Optional[int], optional):
+            Number of tiles to save. Whether they are the first tiles or random is controlled by
+            `random_sample`. If unset, all tiles will be saved. Defaults to None.
+        random_sample (bool, optional):
+            If `save_n_tiles` is set, should the tiles be randomly sampled rather than taken from the
+            beginning of the dataloader. Defaults to False.
+    """
+
```
russelldj, line 70: Even though this function is very easy read, some comments might be helpful, especially since this might be the first part of the codebase someone looks at.

tree_detection_framework/entrypoints/tile_data.py
```diff
@@ -1 +1,201 @@
...
+            save_folder=save_folder,
+            n_tiles=save_n_tiles,
+            random_sample=random_sample,
+        )
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Chipping orthomosaic images")
+
+    parser.add_argument(
```
russelldj, line 99: The explicit help text is probably the nicest option for the user but it ends up being extra effort for the developer and could get out of date if the meaning of any arguments change. In geograypher, I chose to just print the docstring of the function as the argparse description (example [here](https://github.com/open-forest-observatory/geograypher/blob/2900ede9a00ac8bdce22c43e4abb6d74876390f6/geograypher/entrypoints/aggregate_images.py#L191)). Completely fine to leave this as-is since you already did the work to set it up, but this is an option for the future.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -0,0 +1,141 @@
...
+            sample (dict):
+                A dictionary containing the tile to plot. The 'image' key should have a tensor of shape (C, H, W).
+
+        Returns:
+            matplotlib.figure.Figure:
+                A figure containing the plotted image.
+        """
+        # Reorder and rescale the image
+        image = sample["image"].permute(1, 2, 0)
+        image = image.byte().numpy()
```
russelldj, line 42: What does `.byte()` do here?

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -0,0 +1,141 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import rasterio
+import shapely.geometry
+from shapely.affinity import affine_transform
+from torchgeo.datasets import RasterDataset, VectorDataset
+from torchgeo.datasets.utils import BoundingBox, array_to_tensor
+
+
+class CustomRasterDataset(RasterDataset):
```
russelldj, line 14: Since the `plot` method is only applicable for RGB data, maybe we call this class something to indicate that? Or something to indicate it's for orthomosaics specifically?

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class CustomVectorDataset(VectorDataset):
+    """
+    Custom dataset class for vector data which act as labels for the raster data. This class extends the `VectorDataset` from `torchgeo`.
+    """
+
+    def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
+        """Retrieve image/mask and metadata indexed by query.
+           This function is largely based on the `__getitem__` method from torchgeo's `VectorDataset`, with custom modifications for this implementation.
```
russelldj, line 57: Might be useful to explain what additional functionality is being added here.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -76,17 +91,104 @@ def create_dataloader(
...
+
+    for i in tile_indices:
+        sample_bbox = list_of_bboxes[i]
+
+        # Get the referenced sample from the dataloader
+        sample = dataloader.dataset[sample_bbox]
+
+        # Plot the sample image. If the dataloader has label data, index the first dataset to plot.
+        if isinstance(dataloader.dataset, IntersectionDataset):
+            dataloader.dataset.datasets[0].plot(sample)
```
russelldj, line 187: Good logic for now. In the future, is there a way for us to implement a plot method in the interesection dataset? It would be helpful to show both the raster and vector data side by side.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -95,18 +197,69 @@ def save_dataloader_contents(
...
 ):
-    """Save contents of the dataloader to a folder
+    """Save contents of the dataloader to a folder.
 
     Args:
         dataloader (DataLoader):
-            Dataloader to save the contents of
-        save_folder (PATH_TYPE):
+            Dataloader to save the contents of.
+        save_folder (Path):
```
russelldj, line 205: Is there a reason for this change? On line 215 you handle the case where it is a string.

russelldj (approved):
I think this is a really good way to structure this. Other than that, it's just minor comments.

tree_detection_framework/entrypoints/tile_data.py
amrithasp02: I see, I haven't come across this before. Will check it out, thanks.

youngdjn: This sounds smart for the future!

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: It is the PyTorch method to cast a tensor to `torch.uint8` type. The required pixel range for images. I think you had suggested to use this in my first PR.

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: Sure, we can do that. Is there a way I can make the plot function accommodate other types of data?

tree_detection_framework/preprocessing/preprocessing.py
amrithasp02: That's a great point. We can also probably display the crowns draw on the tiles.

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: Ok sounds good. I didn't know about/did remember `byte()` as a convenience method for that.

tree_detection_framework/preprocessing/preprocessing.py
amrithasp02: You're right, I'll change it back.

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: I see two options here. One is to do checks within the `plot` method to determine how many bands there are and visualize appropriately from there. Another option would be to just implement several classes of `Datasets` and use the appropriate ones for each setting. This would actually be my vote since I suspect there are other situations where you'd want to handle different types of data differently, such as preprocessing transformations.

tree_detection_framework/preprocessing/preprocessing.py
russelldj: Really smart. I didn't think of that but I think it's better than side by side.

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: Sounds good

youngdjn (approved):
On a high level this looks good to me! I defer to David on the detailed comments. I think using `utils/geospatial.py` is good.

tree_detection_framework/preprocessing/preprocessing.py
youngdjn: Totally!

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: I think it's reasonable to skip this improvement for the time being and open an issue to do it later if you want.

PR #18, opened 2024-10-14
Implement `RegionDetections` class

russelldj:
This slightly modifies the initial interface for the region detections class and provides and initial implementation of the methods. The goal is to handle the majority of the boilerplate code in the class and have the user specify the simplest objects. 

It also adds a convenience function to convert from a numpy transform matrix to the format expected by shapely.

### DIFF
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import List, Optional, Union
 
 import geopandas as gpd
@@ -7,6 +8,7 @@ import shapely
 from shapely.affinity import affine_transform
 
 from tree_detection_framework.constants import PATH_TYPE
+from tree_detection_framework.utils.geometric import get_shapely_transform_from_matrix
 
 
 class RegionDetections:
@@ -47,15 +49,10 @@ class RegionDetections:
 
         # Check if a pixel to CRS transform is provided
         if pixel_to_CRS_transform:
-            # Format the transform in the format expected by shapely: [a, b, d, e, xoff, y_off]
-            shapely_transform = [
-                pixel_to_CRS_transform[0, 0],
-                pixel_to_CRS_transform[0, 1],
-                pixel_to_CRS_transform[1, 0],
-                pixel_to_CRS_transform[1, 1],
-                pixel_to_CRS_transform[0, 2],
-                pixel_to_CRS_transform[1, 2],
-            ]
+            # Convert from the matrix representation to what is expected by shapely
+            shapely_transform = get_shapely_transform_from_matrix(
+                pixel_to_CRS_transform
+            )
             # Apply this transformation to each geometry to get the detections in a given CRS
             detection_geometries = [
                 affine_transform(geom=geom, matrix=shapely_transform)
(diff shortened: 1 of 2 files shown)


tree_detection_framework/detection/region_detections.py
```diff
@@ -45,51 +81,110 @@ def save(self, save_path: PATH_TYPE):
...
             as_pixels (Optional[bool], optional):
                 Whether to return the values in pixel coordinates. Defaults to False.
 
         Returns:
-            gpd.GeoDataFrame: Detections in the requested CRS
+            gpd.GeoDataFrame: Detections in the requested CRS or in pixel coordinates with a None .crs
         """
-        raise NotImplementedError()
+        # If the data is requested in pixel coordinates, transform it appropriately
+        if as_pixels:
```
amrithasp02, line 107: Is there a specific reason why we can't use the same logic implemented in `__getitem__` method of `CustomVectorDataset` for inverting the transform to convert geospatial coordinates to pixel values? Since both seem to involve similar transformations (if my understanding is correct), I was just curious to know why this method was selected for this context.

amrithasp02 (approved):
Looks really good to me.

tree_detection_framework/detection/region_detections.py
russelldj: I'm now storing the transform directly as a `rasterio.transform`. This cleaned up the logic a lot, good suggestion.

amrithasp02: Awesome

tree_detection_framework/detection/region_detections.py
```diff
@@ -1,41 +1,129 @@
...
 class RegionDetections:
+    detections: gpd.GeoDataFrame
+    pixel_to_CRS_transform: rasterio.transform.AffineTransformer
+    prediction_bounds_in_CRS: Union[shapely.Polygon, shapely.MultiPolygon, None]
+
     def __init__(
         self,
-        detections: gpd.GeoDataFrame,
-        pixel_to_CRS_transform: Optional[np.array] = None,
+        detection_geometries: List[shapely.Geometry],
```
youngdjn, line 21: Is there a case where it would make sense to pass the constructor a GeoDataFrame directly? Is that too unlikely to accommodate?

tree_detection_framework/detection/region_detections.py
```diff
@@ -45,51 +133,102 @@ def save(self, save_path: PATH_TYPE):
...
+        save_path = Path(save_path)
+        save_path.mkdir(parents=True, exist_ok=True)
 
-
-        Returns:
-            gpd.GeoDataFrame: Dataframe of detections with the CRS=self.CRS.
-        """
-        raise NotImplementedError()
+        # Save the detections to a file. Note that the bounds and the CRS are currently lost.
+        self.detections.to_file(save_path)
```
youngdjn, line 141: Why not just store the CRS in the detections GeoDataFrame object using existing gpd functionality? Then it would get saved here.

tree_detection_framework/detection/region_detections.py
youngdjn: Nice!

tree_detection_framework/detection/region_detections.py
```diff
@@ -45,51 +133,102 @@ def save(self, save_path: PATH_TYPE):
...
+            # this will error out
+            detections_in_new_CRS = self.detections.copy().to_crs(CRS)
+            return detections_in_new_CRS
 
 
 class RegionDetectionsSet:
+    region_detections: List[RegionDetections]
+
     def __init__(self, region_detections: List[RegionDetections]):
         """Create a set of detections to conveniently perform operations on all of them
```
youngdjn, line 199: Add something like: "as a single dataframe with an additional column indicating which region each detection belongs too"

tree_detection_framework/detection/region_detections.py
```diff
@@ -45,51 +133,102 @@ def save(self, save_path: PATH_TYPE):
...
+                from using a zero-indexed integer. Defaults to "region_ID".
         """
-        raise NotImplementedError()
+        # Get the detections from each region detection object as geodataframes
+        detection_geodataframes = [rd.get_detections() for rd in self.region_detections]
+
+        # Add a column to each geodataframe identifying which region detection object it came from
+        # Note that dataframes in the original list are updated
+        for i, gdf in enumerate(detection_geodataframes):
+            gdf[region_ID_key] = i
```
youngdjn, line 224: Will it always be sufficient to make this a zero-indexed integer? What about passing through a region ID from the chipping step? Maybe this would be useful so that we can use a single set of IDs across train, val, and test (that are not duplicated for each)? Or maybe that would never be needed. This could be saved for later as an issue (if at all) since it would require some upstream mods too.

youngdjn (approved):
Looks good! Just some naive python questions.

tree_detection_framework/detection/region_detections.py
```diff
@@ -1,41 +1,129 @@
...
         Args:
-            detections (gpd.GeoDataFrame):
-                The dataframe of detected shapes, potentially with additional attributes such as
-                class or confidence. If no CRS is set, the data will be assumed to be in pixel
+            detection_geometries (List[shapely.Geometry]):
+                A list of shapely geometries for each detection. The coordinates can either be
+                provided in pixel coordinates or in the coordinates of a CRS.
+            attributes (Optional[dict], optional):
+                A dictionary mapping from str names for an attribute to a list of values for that
+                attribute, one value per detection. Defaults to {}.
```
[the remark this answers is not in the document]
russelldj: Smart, it might be worth renaming it to `data` since that's what it is called in the dataframe constructor.

tree_detection_framework/detection/region_detections.py
russelldj: Good catch. This comment is actually outdated and the current code works as you suggested. You still loose the bounds and the transform to pixel coordinates though.

tree_detection_framework/detection/region_detections.py
russelldj: I think it would be worth thinking about extending this, but I agree it might be a little complicated so we should save it for a future issue.

PR #27, opened 2024-10-22
Extend `RegionDetection/Set` and implement NMS

russelldj:
Most of this PR is making changes/extensions to the `RegionDetections` and `RegionDetectionsSet` classes to support streamlined NMS. 

- Add an explicit `merge` method to `RegionDetectionSet` to handle the logic of merging multiple regions together or returning a list of per-region detections.
- Rename `get_detections` to `get_data_frame`
- Add a method to get a copy of a `RegionDetections` with a subset of detections.
- Make the instantiation of a `RegionDetections` more flexible so inputs can be provided in geospatial coordinates but also have a transform to pixels associated.

Also I currently have un-used functionality to get the bounds of the predicted region in different CRS for both the `RegionDetections` and `RegionDetectionSet`.

### DIFF
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -319,8 +319,9 @@ class RegionDetectionsSet:
         concatenated_geodataframes = pd.concat(detection_geodataframes)
 
         ## Merge_bounds
-        # get the bounds from each region
-        # Convert to a single shapely object as expected
+        # Get the merged bounds
+        merged_bounds = self.get_bounds(CRS=CRS)
+        # Convert to a single shapely object
         merged_bounds_shapely = merged_bounds.geometry[0]
 
         # Use the geometry column of the concatenated dataframes
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -373,9 +373,9 @@ class RegionDetectionsSet:
     def get_bounds(self, CRS: Optional[pyproj.CRS] = None) -> gpd.GeoSeries:
         region_bounds = [rd.get_bounds(CRS=CRS) for rd in self.region_detections]
         # Create a geodataframe out of these region bounds
-        all_region_bounds = gpd.GeoDataFrame(pd.concat(region_bounds), crs=CRS)
+        all_region_bounds = gpd.GeoSeries(pd.concat(region_bounds), crs=CRS)
         # Compute the union of all bounds
-        merged_bounds = all_region_bounds.geometry.union_all()
+        merged_bounds = gpd.GeoSeries([all_region_bounds.geometry.union_all()], crs=CRS)
 
         return merged_bounds
 
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -437,6 +437,27 @@ class RegionDetectionsSet:
 
         return merged_bounds
 
+    def disjoint_bounds(self) -> bool:
+        """Determine whether the bounds of the sub-regions are disjoint
+
+        Returns:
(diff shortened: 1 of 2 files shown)


russelldj:
I also implemented NMS, which was fairly straightforward. One major limitation is this library only supports rectangular bounding boxes. Currently, this is defined as the minimum bounding rectangle for each polygon (`.bounds`). In the future we might want to support real polygon NMS. Also, there's an option to include a multi-class variation in `LSNMS`, the library we use. I didn't implement it, but we could consider it in the future.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -1,32 +1,94 @@
...
 
     Returns:
         RegionDetections:
             NMS-suppressed set of detections, merged together for the set of regions.
     """
-    # This may implement more sophisticated algorithms, such as down-weighting predictions at the
-    # boundaries or first performing within-tile NMS before across-tile NMS for computational reasons.
-    raise NotImplementedError()
+    # Determine whether to run NMS individually on each region.
+    if run_per_region_NMS:
```
amrithasp02, line 70: This seems like a good approach to eliminate intra-region redundant detections as well as to handle overlapping boundary detections.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -1,32 +1,94 @@
...
+                )
+                for region_detections in detections.region_detections
+            ]
+        )
+        # Merge all the detections into one RegionDetections
+        merged_detections = region_detection_set_NMS_suppressed.merge()
+    else:
+        # Just merge the detections
+        merged_detections = detections.merge()
+    # Run NMS on this merged RegionDetections
```
amrithasp02, line 87: If `run_per_region_NMS` is `True`, I understand that we are essentially doing NMS twice - once for each  sub-region, and the second time for `merged_detections` all together. Just a thought - Is it possible for the sub-regions in a `RegionDetections` to not have overlaps? And if there are no or very little overlaps do you think the second pass of NMS after the merge might be redundant? Especially for larger datasets, I was wondering if it is computationally expensive.

amrithasp02 (approved):
Looks great!

tree_detection_framework/postprocessing/postprocessing.py
russelldj: Good point. I was only thinking of situations where you did tiled inference with an overlapping stride. But it's completely possible that what you suggest could occur. The most likely thing I could think of is some sort of ROI that has multiple disjoint polygons. If your tiles are bigger than the polygons, they might not overlap. Another would be where you do some type of hierarchical NMS on bigger and bigger regions. Then at some point, the regions might become disjoint. 

Since we maintain the `bounds` of each region, we should be able to easily check if there is no overlap. I'll try to implement this.

amrithasp02: I see, that makes sense.

tree_detection_framework/detection/region_detections.py
```diff
@@ -268,35 +342,143 @@ def get_detections(
...
+        # If the union is not requested, return the individual bounds
+        if not union_bounds:
+            return all_region_bounds
+
+        # Compute the union of all bounds
+        merged_bounds = gpd.GeoSeries([all_region_bounds.geometry.union_all()], crs=CRS)
+
+        return merged_bounds
+
+    def disjoint_bounds(self) -> bool:
```
russelldj, line 440: @amrithasp02 when you get a chance, can you do a quick check to make sure that the reasoning here is sound?

amrithasp02: Yes, it looks good!

PR #32, opened 2024-10-28
Add `plot` method to both `RegionDetections` and `RegionDetectionSet`

russelldj:
This handles a lot of the logic around plotting while relying heavily on the functionality from `geopandas`. I also added an example notebook showing the visualizations.


amrithasp02 (approved):
Very cool to see NMS in action

Issue #26, opened 2024-10-21
Allow the user to specify whether `CustomRasterDataset` returns float or uint8 data


russelldj:
Some approaches expect one versus the other. This is a reversal of my previous logic that the dataloaders should be standardized and the downstream steps should handle conversions. I think this proposed option will reduce boilerplate and also take advantage of the multi-threaded pre-processing implemented in the dataloader.

russelldj:
Make this a low priority and just set the default to whatever is expected by DeepForest. This might require doing some sort of transform.

PR #34, opened 2024-10-29
Implementing DeepForest with Pytorch Lightning

amrithasp02:
This PR focuses on fitting DeepForest into the framework including - training and predicting.

Major changes to look out for:

1. `preprocessing.py` has a `CustomDataModule` class which inherits from `LightningDataModule`. I was unable to directly use the DataLoaders created by `preprocessing.create_dataloader` for lightning since it expects a `LightningdataModule`.
2. I created a derived class `DeepForestDetector` from `LightningDetector` following our initial logic. Also included `DeepForestModule` as a new class.
3. `training_step()` in `DeepForestModule` had to be heavily modified to make sure the images and labels are in the format expected by `RetinaNet` (backbone for deepforest).
4. `predict()` method in `DeepForestDetector` : I decided to directly use `deepforest` library's method for this to avoid unnecessary replication of code and rewriting.
5. Included 3 staticmethods (defined by David) to help post-process the prediction outputs to `RegionDetectionsSet` objects. 
6. Included a notebook `deepforest_train_and_predict.ipynb` for demonstration.


tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+            weights=RetinaNet_ResNet50_FPN_Weights.COCO_V1
+        )
+
+        return backbone
+
+    def create_anchor_generator(
+        self, sizes=((8, 16, 32, 64, 128, 256, 400),), aspect_ratios=((0.5, 1.0, 2.0),)
+    ):
+        """
+        Create anchor box generator as a function of sizes and aspect ratios
```
russelldj, line 183: If a method/function directly calls another one, I'll often omit the documentation except for for indicating which function is being called and saying to refer to its documentation. Partially to conserve space, partially to avoid having outdated documentation if the called function is updated.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+
+        """
+        anchor_generator = AnchorGenerator(sizes=sizes, aspect_ratios=aspect_ratios)
+
+        return anchor_generator
+
+    def create_model(self):
+        """Create a retinanet model
+        Args:
+            num_classes (int): number of classes in the model
```
russelldj, line 204: It looks like it no longer takes args.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+            nms_thresh (float): non-max suppression threshold for intersection-over-union [0,1]
+            score_thresh (float): minimum prediction score to keep during prediction  [0,1]
+        Returns:
+            model: a pytorch nn module
+        """
+        resnet = self.load_backbone()
+        backbone = resnet.backbone
+
+        model = RetinaNet(backbone=backbone, num_classes=self.param_dict["num_classes"])
+        # model.nms_thresh = self.param_dict["nms_thresh"]
```
russelldj, line 214: Can you remove these lines or turn them into TODOs/comments if needed? Also we should think about how we want to handle NMS since that is duplicated downstream. But that doesn't need to be now.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+                During training, it returns a dict[Tensor] which contains the losses.
+        """
+        return self.model.forward(images, targets)  # Model specific forward
+
+    def training_step(self, batch):
+        # Ensure model is in train mode
+        self.model.train()
+        device = next(self.model.parameters()).device
+
+        # Image is expected to be a list of tensors, each of shape [C, H, W] in 0-1 range.
```
russelldj, line 270: At some point we should consider whether we want to replace this with #26 . But for now it's completely reasonable.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+        image_batch = (batch["image"][:, :3, :, :] / 255.0).to(device)
+        image_batch_list = [image_batch[i] for i in range(image_batch.size(0))]
```
russelldj, line 272: I think you can do `[x for x in image_batch]` as a more succinct equivalent method if you want. At least for numpy arrays this will iterate over the first dimension.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+
+        # To store every image's target - a dictionary containing `boxes` and `labels`
+        targets = []
+        for tile in batch["bounding_boxes"]:
+            # Convert from list to FloatTensor[N, 4]
+            boxes_tensor = torch.tensor(tile, dtype=torch.float32).to(device)
+            # Need to remove boxes that go out-of-bounds. Has negative values.
+            valid_mask = (boxes_tensor >= 0).all(dim=1)
```
russelldj, line 280: Do ones that extend beyond the bounds in the positive direction have this issue to? Again, fine for now but we should consider how to handle this in pre-processing as a longer-term solution.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+            factor=0.1,
+            patience=10,
+            verbose=True,
+            threshold=0.0001,
+            threshold_mode="rel",
+            cooldown=0,
+            min_lr=0,
+            eps=1e-08,
+        )
+        # # Monitor rate is val data is used
```
russelldj, line 313: Can you remove this or turn it into a more descriptive TODO?

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+
+        Returns:
+            `RegionDetectionsSet`: A collection of region detection objects, where each detection object contains:
+            - Predicted bounding geometries (in shapely format) for detected regions
+            - Associated prediction dataframe
+            - Pixel-based and geospatial bounding boxes for each detection
+            - CRS information
+        """
+        # Load deepforest model
+        model = main.deepforest()
```
russelldj, line 379: I think this should just be `self.model`?

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+        # Store all batched elemts from the dataloader in a list
+        all_tiles = list(inference_dataloader)
+        logging.info("Getting predictions from deepforest...")
+        output_dataframes = []
+        for tile in all_tiles:
+            image = tile["image"][0]  # assumes batch_size 1
+            # Convert to datatype expected by deepforest
+            image_np = image.numpy().astype(np.float32)
+            # Retain first 3 channels and reorder it so that RGB channels are the last dimension
+            image_np = image_np[:3, :, :].transpose(1, 2, 0)
```
russelldj, line 393: Yet another thing we might want to consider handling in pre-processing in the future.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+        model = main.deepforest()
+        # Get the latest model weights
+        # TODO: Support loading a different pretrained model from checkpoint
+        model.use_release()
+
+        # Store all batched elemts from the dataloader in a list
+        all_tiles = list(inference_dataloader)
+        logging.info("Getting predictions from deepforest...")
+        output_dataframes = []
+        for tile in all_tiles:
```
russelldj, line 388: I think it's actually pretty important that we support non-singleton batches here. Do you think that would be possible?

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+                geospatial_prediction_bounds=geospatial_bounds,
+                input_in_pixels=True,
+                CRS=CRS,
+            )
+            # Append to the list
+            region_detections.append(region_detection)
+
+        logging.info("Done.")
+        # Return the region detection set
+        return RegionDetectionsSet(region_detections)
```
russelldj, line 427: In the future we may want to make this a generator and return `RegionDetections` but this is good for now.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+
+    def train(
+        self,
+        model: DeepForestModule,
```
russelldj, line 431: Do we need to pass in the model here? I was thinking we'd just use `self.model`.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+                CRS=CRS,
+            )
+            # Append to the list
+            region_detections.append(region_detection)
+
+        logging.info("Done.")
+        # Return the region detection set
+        return RegionDetectionsSet(region_detections)
+
+    def train(
```
russelldj, line 429: Could this functionality be moved into `LightningDetector`? It doesn't seem specific to `DeepForestDetector`

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -279,3 +284,49 @@ def save_dataloader_contents(
...
+        # create the data based on the stage the Trainer is in
+        if stage == "fit":
+            self.train_data = self.raster_data & self.vector_data  # IntersectionDataset
+        if stage == "validate" or stage == "fit":
+            self.val_data = self.raster_data & self.vector_data
+        if stage == "test":
+            self.test_data = self.raster_data & self.vector_data
+
+    def train_dataloader(self):
+        sampler = GridGeoSampler(self.train_data, size=350, stride=400)
```
russelldj, line 320: Can we provide the size and stride as class attributes?

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -279,3 +284,49 @@ def save_dataloader_contents(
             break
 
     print(f"Saved {saved_tiles_count} tiles to {save_folder}")
+
+
+class CustomDataModule(GeoDataModule):
```
russelldj, line 289: This is really cool, I'd like to hear more about this concept

tree_detection_framework/utils/detection.py
```diff
@@ -0,0 +1,101 @@
+import json
+import os
+import urllib
+
+import pandas as pd
+from tqdm import tqdm
+
+
+class DownloadProgressBar(tqdm):
```
russelldj, line 9: If this is defined elsewhere can you import it from there?

tree_detection_framework/utils/detection.py
```diff
@@ -0,0 +1,101 @@
...
+        Returns:
+
+        """
+        if tsize is not None:
+            self.total = tsize
+        self.update(b * bsize - self.n)
+
+
+def use_release_df(
+    save_dir=os.path.join(os.path.abspath(os.path.dirname(__file__)), "data/"),
```
russelldj, line 29: You could make this a constant defined in `constants.py` if you want to make it clearer what it represents.

tree_detection_framework/detection/detector.py
```diff
@@ -88,3 +121,355 @@ def save_model(self, save_file: PATH_TYPE):
...
+                xmin=tile_bounds.minx,
+                ymin=tile_bounds.miny,
+                xmax=tile_bounds.maxx,
+                ymax=tile_bounds.maxy,
+            )
+            for tile_bounds in batch_bounds
+        ]
+
+
+class RetinaNetModel:
```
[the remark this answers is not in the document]
amrithasp02: Yep, we should. Shall I create a file called `backbone.py` in the `tree_detection_framework/detection` folder and add it to that? Let me know if you feel there is a better name or location.

tree_detection_framework/detection/detector.py
amrithasp02: Description of params expected by retinanet ([link](https://pytorch.org/vision/main/models/generated/torchvision.models.detection.retinanet_resnet50_fpn.html)) states that bounding box values should fall in this range:  0 <= x1 < x2 <= W and 0 <= y1 < y2 <= H

So I decided to remove boxes that go outside image bounds, just to be safe. Although, I haven't tested if it breaks when included.

tree_detection_framework/detection/detector.py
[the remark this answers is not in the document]
russelldj: I don't have a strong preference. My initially thought was both the "backbones" and LightningModules could both go in a file called `tree_detection_framework/detection/models.py`. But I'll leave it up to your judgement.

tree_detection_framework/detection/detector.py
russelldj: That makes sense. It just looked like you only checked for negative values and not for exceeding the width or height.

amrithasp02: I just re-read your question, and you are right. I am checking only for negative values and not for bounds extending in positive direction. But it hasn't shown any errors while training.

tree_detection_framework/detection/detector.py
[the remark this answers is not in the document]
amrithasp02: That sounds like a better option. This way we can have only the "Detectors" in `detector.py`

tree_detection_framework/detection/detector.py
amrithasp02: I was initially waiting for us to be done with the `detectree2` implementation and then make the decision on that based on if the `train()` method looks the same for both. But I think it should most likely be. I'' move it to `LightningDetector` then.

tree_detection_framework/detection/detector.py
amrithasp02: Since right now we are calling `setup_model()` only if we want to train (and not calling `setup_model()` when we want to just get predictions - as it uses `deepforest` library's model directly) I had to pass model here.

tree_detection_framework/detection/detector.py
russelldj: Actually yeah I think it would help to talk about this at the meeting today. I initially thought you could call `self.model.forward(image)` to run inference. If you look at the source code for [RetinaNet](https://github.com/pytorch/vision/blob/e9a3213524a0abd609ac7330cf170b9e19917d39/torchvision/models/detection/retinanet.py#L573), the `targets` parameter is optional, and if it isn't specified, the model returns only predictions. So I think you could make `targets` optional in the `DeepForestModule` as well and then you could use it either for training or inference.

tree_detection_framework/detection/detector.py
russelldj: I think you might be able to avoid the trainer entirely. Since the RetinaNet [forward](https://github.com/pytorch/vision/blob/e9a3213524a0abd609ac7330cf170b9e19917d39/torchvision/models/detection/retinanet.py#L573) accepts a list of Tensors, I think as long as you provided your input in that format you could call `forward` directly.

tree_detection_framework/preprocessing/preprocessing.py
amrithasp02, line 289: @russelldj do you think I should move the `CustomDataModule` class to `tree_detection_framework/preprocessing/derived_geodatasets.py`?

russelldj: Yeah sounds like a good idea

russelldj (approved):
Great, work it all looks good to me

PR #36, opened 2024-10-30
Implement random detector and add convenience methods to `Detector` class

russelldj:
The majority of the work on this PR is to add/modify a few methods for the `Detector` class.
-  Rather than `predict` being implemented by each derived class, now `predict_batch` is, only for a single batch.
- The `predict_as_generator` method can be used to generate predictions sequentially from a dataloader. This method handles most of the boilerplate functionality.
- The `predict` method now returns either a list of `RegionDetections` or a single `RegionDetectionSet`. 

The random detector is now fairly simple. It generates a fixed number of detections of a given size for each image. Additional sophistication could be added (e.g. a range of sizes, or a specified distribution of spatial locations) but I didn't see an immediate need for this functionality.


tree_detection_framework/detection/detector.py
```diff
@@ -15,21 +20,210 @@ def setup(self):
...
 
         Args:
             inference_dataloader (DataLoader): Dataloader to generate predictions for
+        """
+        # Iterate over each batch in the dataloader
+        for batch in inference_dataloader:
+            # This is the expensive step, generate the predictions using predict_batch from the
+            # derived class. The additional arguments are also passed to this method with kwargs
+            batch_preds_geometries, batch_preds_data = self.predict_batch(
+                batch, **kwargs
```
amrithasp02, line 38: Does this mean I should name the predict method I will be adding in `DeepForestDetector` class which is supposed to use `model.forward()` to get the batch predictions as `predict_batch()`?

tree_detection_framework/detection/detector.py
```diff
@@ -15,21 +20,210 @@ def setup(self):
...
+
+
+class RandomDetector(Detector):
+    """A detector that produces random detections primarily used for testing"""
+
+    def predict_batch(
+        self,
+        batch: dict,
+        detections_per_tile: int = 10,
+        detection_size_fraction: float = 0.1,
```
amrithasp02, line 174: Should we have a check to ensure that `detection_size_fraction` is between 0 and 1?

amrithasp02 (approved):
Looks good to me.

tree_detection_framework/detection/detector.py
russelldj: Yeah if you're willing to modify it to be in this format I think that would be a good idea. You'll have to make some changes so it's in the expected structure. Instead of generating predictions on a whole dataloader, you'll now only do it for a single batch from the dataloader. And you'll have to set up the returns so it matches the expected format, with the first return being the per-sample geometries and the second being the per-sample additional attribute, such as confidence.

PR #37, opened 2024-11-01
Add mask-to-shapely function and example

### LINKED ISSUE #30: Add a function to convert a binary mask to a `shapely.Polygon/MultiPolygon`
This function should take in a single binary mask and produce a polygon. It should accommodate multiple disconnected regions and holes within regions.

russelldj:
This adds a function to convert a binary mask to a shapely multipolygon and closes #30. It's inspired by @amrithasp02 's initial work using matplotlib and directly calls `ContourPy`, the library used by matplotlib. In the future we could explore other library implementations, but for now this seems relatively optimized and convenient. The function largely follows an example [here](https://contourpy.readthedocs.io/en/v1.3.0/user_guide/external/shapely.html#filled-contours-to-shapely).

I also added an example of using this function to visualize the boundaries of the predictions over the images. While it is not included in the tiles that are shown, I confirmed that this approach correctly handles both holes and multiple disconnected regions from the same mask. In both cases, all the information is represented in one `shapely.MultiPolygon`. As discussed, we could later post-process this to do things such as retaining the largest single component or filling holes.

### DIFF
--- a/tree_detection_framework/utils/geometric.py
+++ b/tree_detection_framework/utils/geometric.py
@@ -1,4 +1,6 @@
 import numpy as np
+import shapely
+from contourpy import contour_generator
 
 
 def get_shapely_transform_from_matrix(matrix_transform: np.ndarray):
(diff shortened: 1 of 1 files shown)


amrithasp02 (approved):
Looks good!

PR #40, opened 2024-11-05
Add entrypoint script to use deepforest for prediction

amrithasp02:
This PR deals with adding an entrypoint script that creates a dataloader and gets predictions from a tree detection model specified by the user. For now, we have `DeepForest` as the model.


tree_detection_framework/entrypoints/generate_predictions.py
```diff
@@ -1 +1,87 @@
...
+
+    if save_folder:
+        outputs.save(save_folder)
+
+    if view_predictions_plot is True:
+        outputs.plot()
+
+
+def parse_args() -> argparse.Namespace:
+    description = (
```
russelldj, line 62: How do you feel about [this style](https://github.com/open-forest-observatory/geograypher/blob/2900ede9a00ac8bdce22c43e4abb6d74876390f6/geograypher/entrypoints/label_polygons.py#L121) of documentation, where the function has a comprehensive docstring and the argparse description just prints that out? I'm slightly in favor of it because it allows the function to be documented so it can be used easily elsewhere, but I know it's clunky.

tree_detection_framework/entrypoints/generate_predictions.py
```diff
@@ -1 +1,87 @@
...
+    if tree_detection_model == "deepforest":
+
+        # Setup the parameters dictionary
+        param_dict = {
+            "backbone": "retinanet",
+            "num_classes": 1,
+        }
+
+        df_module = DeepForestModule(param_dict)
+        lightning_detector = DeepForestDetector(df_module)
```
russelldj, line 50: Can you provide a descriptive error if the `tree_detection_model` is not one of the supported ones?

tree_detection_framework/entrypoints/generate_predictions.py
```diff
@@ -1 +1,87 @@
...
+
+    outputs = lightning_detector.predict(dataloader)
+
+    if save_folder:
+        outputs.save(save_folder)
```
russelldj, line 55: The `RegionDetectionsSet.save(*)` method takes as input a path to a geofile, not a folder.

tree_detection_framework/entrypoints/generate_predictions.py
```diff
@@ -1 +1,87 @@
...
+        # Setup the parameters dictionary
+        param_dict = {
+            "backbone": "retinanet",
+            "num_classes": 1,
+        }
+
+        df_module = DeepForestModule(param_dict)
+        lightning_detector = DeepForestDetector(df_module)
+
+    outputs = lightning_detector.predict(dataloader)
```
russelldj, line 52: I'll often comment on which line is the expensive step if it's not immediately obvious.

russelldj (approved):
Nice work, just some additions.

tree_detection_framework/constants.py
```diff
@@ -2,11 +2,11 @@
 from typing import Union
 
 import geopandas as gpd
-import numpy as np
+import numpy.typing
```
russelldj, line 5: Good fix, I saw this was an issue and didn't look into it.

youngdjn:
Nice, I'm going to try running this today, so @amrithasp02, please wait for my review before merging.

tree_detection_framework/entrypoints/generate_predictions.py
amrithasp02: Ohh I see, so the input path must end with `.gpkg` ? I did notice that when I tested it it with a folder path it saved 5 files with different extensions that I could not understand

tree_detection_framework/constants.py
amrithasp02: I thought it was an issue after I installed `lsnms` so I made a whole new environment and this error still showed up. So ended up changing it.

russelldj:
There's an example of setting the log level here: https://github.com/open-forest-observatory/geograypher/blob/2900ede9a00ac8bdce22c43e4abb6d74876390f6/geograypher/entrypoints/orthomosaic_predictions.py#L79

youngdjn (approved):

PR #41, opened 2024-11-07
Add Jupyter compatibility

youngdjn:
Including `ipykernel` in the conda env enables this env to be discovered and used as a kernel for Jupyter notebooks. I'm not sure if Jupyter compatibility is enough of a priority to include this extra dependency in the default readme installation instructions. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,14 @@ in the base conda environment or with no environment active.
 ```
 curl -sSL https://install.python-poetry.org | python3 -
 ```
-Now create and activate a conda environment for the dependencies of this project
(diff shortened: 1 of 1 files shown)


amrithasp02 (approved):
I personally use Jupyter a lot for interactive data exploration and prototyping, so including `ipykernel` as an environment dependency makes a lot of sense. We could also probably add a note in the README explaining why it's included and mentioning that it can be left out if Jupyter compatibility isn't needed.

youngdjn:
Thanks, I included a note to that effect.

PR #49, opened 2024-11-08
Add score-based filtering and progress bar for predictions

amrithasp02:
Included an option to filter the detections based on confidence scores. I added this as a step prior to NMS within the same function since it seemed easier & straightforward. But let me know if you think having it as a separate function is a better approach. 

Also added the progress bar within the `predict_as_generator` method.

Note: Ignore changes made in `detector.py`. It has a previous commit that I reverted but I have changed the name of a variable. The logic hasn't changed.


amrithasp02:
Now you can include `--iou-threshold` and `--min-confidence` as flags in the `generate_predictions.py` entrypoint script

tree_detection_framework/detection/detector.py
```diff
@@ -44,42 +46,49 @@ def predict_as_generator(
...
-                    CRS=CRS,
-                    input_in_pixels=True,
-                    pixel_prediction_bounds=image_bounds,
-                    geospatial_prediction_bounds=geospatial_bounds,
+        total_batches = len(inference_dataloader)
+        with tqdm(
+            total=total_batches, desc="Performing prediction on batches"
+        ) as batch_pbar:
+            # Iterate over each batch in the dataloader
+            for batch in inference_dataloader:
```
russelldj, line 54: Is the more complicated context manager syntax required to work with a generator? My initial thought would have just been to do `for batch in tqdm(inference_dataloader, desc="Performing prediction on batches")`

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -12,6 +12,7 @@ def single_region_NMS(
     detections: RegionDetections,
     iou_theshold: float = 0.5,
     confidence_column: str = "score",
+    min_confidence: float = 0.3,
```
russelldj, line 15: I think including the min confidence in NMS is convenient and works for now. If needed, we can add a standalone filter later.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -30,6 +33,13 @@ def single_region_NMS(
     # Extract the geodataframe for the detections
     detections_df = detections.get_data_frame()
 
+    # Filter detections based on minimum confidence score
+    detections_df = detections_df[detections_df[confidence_column] >= min_confidence]
+    if detections_df.empty:
+        logging.info("All detections were filtered out by confidence threshold.")
+        # Return empty if no detections pass threshold
+        return detections.subset_detections([])
```
russelldj, line 41: Good catch, though you may get this print out a lot for individual tiles that don't have any trees.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -61,7 +72,8 @@ def multi_region_NMS(
             What intersection over union value to consider an overlapping detection. Defaults to 0.5.
         confidence_column (str, optional):
             Which column in the dataframe to use as a confidence for NMS. Defaults to "score"
-
+        min_confidence (float, optional):
+            Prediction score threshold for detections to be included.
```
russelldj, line 76: Nitpick, can you add the separating line following the args back in?

russelldj (approved):
Looks good, just some questions and optional changes.

tree_detection_framework/detection/detector.py
amrithasp02: Thanks, this simplified the loop!

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: Yep, I have remove it now.

PR #60, opened 2024-11-26
Add a GeometricDetector class

amrithasp02:
This PR focuses on adding a new type of detector to the framework that uses the geometric learning-free approach to produce tree crown detections using a CHM.

Algorithm followed to obtain treetops:
[Seeing the Trees in the Forest: Using Lidar and Multispectral Data Fusion with Local Filtering and Variable Window Size for Estimating Tree Height](https://pdfs.semanticscholar.org/a9e1/9a00286959b2c059ee7678502b0dad9cf639.pdf)

Algorithm followed to produce tree crowns:
[Imputation of individual longleaf pine (Pinus palustris Mill.) tree attributes from field and LiDAR data](https://research.fs.usda.gov/treesearch/52704)

To visualize the predictions, please view `tree-detection-framework/sandbox/detection/geometric_detections.ipynb`


tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+        self, a: float = 0.00901, c: float = 2.52503, res: float = 0.2, min_ht: int = 5
+    ):
+        self.a = a
+        self.c = c
+        self.res = res
+        self.min_ht = min_ht
+
+    def _get_three_polygon_intersection(self, row):
+        # Perform intersections sequentially
+        intersection = (
```
russelldj, line 279: Could you instead use [intersection_all](https://shapely.readthedocs.io/en/2.0.6/reference/shapely.intersection_all.html#shapely.intersection_all)?

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+
+        # Return an empty MultiPolygon if there's no intersection
+        if intersection.is_empty:
+            return MultiPolygon()
+        elif isinstance(intersection, MultiPolygon):
+            # Return directly if the result is already a MultiPolygon
+            return intersection
+        elif intersection.geom_type == "Polygon":
+            # Wrap a single Polygon as a MultiPolygon
+            return MultiPolygon([intersection])
```
russelldj, line 293: Is this required? I think geopandas can handle inputs that are a mix of polygon and multipolygon data.

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+        else:
+            # For unexpected cases
+            return MultiPolygon()
```
russelldj, line 296: This could alternatively be an error or at least a warning. I'm not sure when it would happen though.

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+
+    def get_tree_crowns(self, image) -> List[shapely.MultiPolygon]:
+        """Generate tree crowns for an image.
+
+        Args:
+            batch (dict): A batch from the torchgeo dataloader
```
russelldj, line 302: The argument needs to be updated.

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+                    continue
+
+                # Calculate the radius
+                radius = self.a * (ht**2) + self.c
+                # Convert radius from meters to pixels
+                radius_pixels = radius / self.res
+                side = int(np.ceil(radius_pixels))
+
+                # Define bounds for the neighborhood
+                i_min = max(0, i - side)
```
russelldj, line 335: There are some tricks that could be used to make this faster based off what @youngdjn mentioned a few days ago. Specifically, if we defined the `(region_i - i) ** 2` and `(region_j - j) ** 2` terms as column and row vectors, respectively, that operation would be more efficient. Then the sum could be broadcast. This might remove the need for `meshgrid` in the first place.

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+
+        # Get Voronoi Diagram from the calculated treetop points
+        voronoi_diagram = shapely.voronoi_polygons(MultiPoint(all_treetop_pixel_coords))
+
+        # Store the individual polygons from Voronoi diagram in the same sequence as the treetop points
+        ordered_polygons = []
+        for treetop_point in all_treetop_pixel_coords:
+            for polygon in voronoi_diagram.geoms:
+                # Check if the treetop is inside the polygon
+                if polygon.contains(treetop_point):
```
russelldj, line 366: Is there a way to make sure that the veronie diagram elements are ordered the same way to avoid this step?

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+        # 2. A set of multipolygons geenrated from the binary mask of the image
+        all_radius_in_pixels = []
+        all_circles = []
+        all_multipolygon_masks = []
+
+        for treetop_point, treetop_height in zip(
+            tile_gdf["treetop_pixel_coords"], tile_gdf["treetop_height"]
+        ):
+            # Compute radius as 0.6 times the height, divide by resolution to convert unit to pixels
+            radius = (0.6 * treetop_height) / self.res
```
russelldj, line 390: Can we provide the radius factor as a parameter?

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+            # Convert to an integer rounded to the upper value
+            radius = int(np.ceil(radius))
```
russelldj, line 392: Is there a reason this needs to be an integer?

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+        # Here, we get 2 new sets of polygons:
+        # 1. A circle for every detected treetop
+        # 2. A set of multipolygons geenrated from the binary mask of the image
+        all_radius_in_pixels = []
+        all_circles = []
+        all_multipolygon_masks = []
+
+        for treetop_point, treetop_height in zip(
+            tile_gdf["treetop_pixel_coords"], tile_gdf["treetop_height"]
+        ):
```
russelldj, line 388: There should be vectorized ways to make all of the circles at once using `gpd` objects. This isn't required though.

tree_detection_framework/detection/detector.py
```diff
@@ -262,6 +264,180 @@ def predict_batch(
...
+            radius = (0.6 * treetop_height) / self.res
+            # Convert to an integer rounded to the upper value
+            radius = int(np.ceil(radius))
+            all_radius_in_pixels.append(radius)
+
+            # Create a circle by buffering it by the radius value and add to list
+            all_circles.append(treetop_point.buffer(radius))
+
+            # Calculate threshold value for the binary mask as 30% of the treetop height
+            threshold = 0.3 * treetop_height
```
russelldj, line 399: Same with this, can we specify it as a parameter?

russelldj (commented):
This looks good. I think there's a few things that could make it a little more efficient or flexible that I'd prefer to have included, but I'm also ok with it being merged as-is.

youngdjn:
A couple things we discussed in this morning's meeting (not exhaustive):

- Splitting treetop detection and crown delineation into separate methods
- Adding a linear term (coefficient `b`) to the treetop window function
- Vectorizing crown delineation across all rows of the treetops data frame, as opposed to a for loop
- See creating the height mask is more efficient by first cropping the tile CHM to the maximum possible bounds of the tree crown, as opposed to applying the mask to the whole tile CHM for each tree
- If a crown consists of a multipolygon, keep just the polygon in which the treetop point actually falls

tree_detection_framework/detection/detector.py
amrithasp02: Yep, it seems unnecessary to have that, I've simplified it now.

tree_detection_framework/detection/detector.py
amrithasp02: Made the change

tree_detection_framework/detection/detector.py
amrithasp02: I've added it as a TODO task since I could not find a way to do that using this method

tree_detection_framework/detection/detector.py
amrithasp02: You're right, it is not needed. I removed that step.

PR #61, opened 2024-11-28
Added SAM2 into detector framework

mich889:
Added SAM2 into models 
Results are shown here: https://docs.google.com/presentation/d/1ZsbVOOaRL18q__D5IOIF2D-fgIMiABEKhd8ECb8XieA/edit#slide=id.g31ae21ffab1_0_24
To install sam2 follow instructions [here](https://github.com/facebookresearch/sam2)
But essentially you must run
```
git clone https://github.com/facebookresearch/sam2.git && cd sam2

pip install -e .
```
and to install checkpoints:
```
cd checkpoints && \
./download_ckpts.sh && \
cd ..
```

NOTE: iopath versions are incompatible between SAM2 (iopath>=0.1.10) and Detectron2 (iopath<0.1.10,>=0.1.7) which may lead to issues when trying to use them simultaneously in tree detector framework.


tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+            # Get prediction scores
+            scores = [dic["stability_score"] for dic in pred]
+            # unlabelled masks
+            labels = None
+            all_data_dicts.append({"score": scores, "labels": labels})
+
+        return all_geometries, all_data_dicts
+
+
+class SAMV2Detector(Detector):
```
russelldj, line 514: It looks like this class is duplicated, I assume a simple copy/paste error. Is there something I'm missing?

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+    def __init__(self):
+        self.device = (
+            torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
```
russelldj, line 517: This patterns shows up a fair bit, we could could make `DEVICE` or `DEFAULT_DEVICE` a constant in `tree_detection_framework/constants.py`.

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+        )
+
+    def setup_predictor(self):
```
russelldj, line 520: This `setup_predictor` method was probably a bad design choice on our part since it means that the model is created from scratch for each batch. You can leave this as-is, or if you have a way you want to refactor the class to only create the model once you are welcome to do it.

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+        sam2_checkpoint = "checkpoints/sam2.1_hiera_large.pt"
+        model_cfg = "configs/sam2.1/sam2.1_hiera_l.yaml"
```
russelldj, line 522: I think it could be useful to have these as parameters which can be defaulted to the current values. But it's fine as-is.

russelldj: Also, does data need to be copied somewhere for these paths to work out?

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+        """
+        self.setup_predictor()
+
+        with torch.no_grad():
+            masks = []
+            for original_image in batch:
+                original_image = (
+                    original_image.permute(1, 2, 0).byte().numpy()
+                )  # (4, h, w)
+                # permuted_image = original_image.permute(1, 2, 0)
```
russelldj, line 547: Can you please remove the old code?

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+        Returns:
+            masks List[List[Dict]]: list of dictionaries for each mask in the batch
+        """
+        self.setup_predictor()
+
+        with torch.no_grad():
+            masks = []
+            for original_image in batch:
+                original_image = (
+                    original_image.permute(1, 2, 0).byte().numpy()
```
russelldj, line 545: I assume you confirmed the model expects a byte image (0-255) and not something else, such as float 0-1?

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+                )  # (4, h, w)
+                # permuted_image = original_image.permute(1, 2, 0)
+                rgb_image = original_image[:, :, :3]  # (h, w, 3)
```
russelldj, line 548: Subsetting to 3 channels is completely fine, but you could also raise an error if you get a different number of channels. Up to you. As an aside, would this support one-channel data? We have been using one-channel canopy height rasters in this framework as well.

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+                A list of predictions one per image in the batch. The predictions for each image
+                are a list of shapely objects.
+            all_data_dicts (Union[None, List[dict]]):
+                Predicted scores and classes
+        """
+        images = batch["image"]
+
+        self.setup_predictor()
+
+        batch_preds = self.call_predict(images)
```
russelldj, line 574: Even if it's fairly self-explanatory, I often like to leave a comment for where the computationally-expensive part of the function/method is.

tree_detection_framework/detection/detector.py
russelldj: I'll give this a try in a bit, but did you confirm that this properly runs on the GPU? I wonder if you might need a `.to(device)` call in some situations.

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+            shapely_objects = [
+                mask_to_shapely(pred_mask) for pred_mask in segmentations
+            ]
+
+            all_geometries.append(shapely_objects)
+
+            # Get prediction scores
+            scores = [dic["stability_score"] for dic in pred]
+            # unlabelled masks
+            labels = None
```
russelldj, line 597: We don't have a standardized convention around this, but I think the `labels` field could just be omitted. Was there a reason you chose to include it?

russelldj (commented):
Nice work! This is really cool. I think it could be merged as-is if needed, but I had a few questions and comments that could make it better.

russelldj:
It looks like #60 caused a minor conflict. Can you either merge main into your feature branch or rebase onto main to resolve that issue?

russelldj:
Also, thanks for looking into the dependency issues. I created an issue here #62 that you could try to work through if you have time. But just the description you provided is helpful.

russelldj:
Also, if you get a chance, could you add `SAM2` as an option in the example [prediction notebook](https://github.com/open-forest-observatory/tree-detection-framework/blob/main/examples/predict_detections.ipynb)?

tree_detection_framework/detection/detector.py
```diff
@@ -421,6 +422,184 @@ def save_model(self, save_file: PATH_TYPE):
...
+
+    def call_predict(self, batch):
+        """
+        Args:
+            batch (Tensor): 4 dims Tensor with the first dimension having number of images in the batch
+
+        Returns:
+            masks List[List[Dict]]: list of dictionaries for each mask in the batch
+        """
+        self.setup_predictor()
```
amrithasp02, line 539: It seems like `self.setup_predictor()` is called twice - once in `call_predict()` and again in `predict_batch()`. The call to `self.setup_predictor()`  in this method could be removed since `predict_batch()` is called first. Or to simplify things, I think moving this line to `__init__` method would be better so that we can avoid calling it for every batch like @russelldj  had mentioned.

russelldj: Great idea. Taking it one step further, `setup_predictor` could be removed entirely and that code could be moved into `__init__`.

tree_detection_framework/detection/detector.py
russelldj: Trying to run this confirms that it only works if these are valid relative paths to the file. So this code would need to know where `sam2` is installed. You might be able to introspect this from the `sam2` module code (I think you might be able to get the installed path of a module) or you could make the user provide a valid relative or absolute path.

tree_detection_framework/detection/detector.py
mich889: when i tried loading the `original_image` on the gpu i get the error:
`NotImplementedError: Image format not supported`

It seems the sample notebook follows the same formatting where the input to the `generate` function is an numpy array not a tensor

tree_detection_framework/detection/detector.py
mich889: based on [this](https://github.com/facebookresearch/segment-anything/issues/642) issue it seems it only supports 3 channels. in that case is it possible to turn 1 channel to 3 channel (grayscale -> rgb?). i tried condensing the original_image to 1 channel with `original_image[:, :, :1]` and it seemed to error out so im not too sure if its supported but you can give it a try with the canopy height rasters

tree_detection_framework/detection/detector.py
mich889: i updated the README for the download instructions for the model and checkpoint, let me know if it works now!

tree_detection_framework/detection/detector.py
mich889: I was following the formatting from the other detectors but if its unecessary i can also remove

tree_detection_framework/detection/detector.py
```diff
@@ -767,6 +768,91 @@ def save_model(self, save_file: PATH_TYPE):
...
+        """
+
+        with torch.no_grad():
+            masks = []
+            for original_image in batch:
+                if original_image.shape[0] < 3:
+                    raise ValueError("Original image has less than 3 channels")
+
+                original_image = original_image.permute(1, 2, 0).byte().numpy()
+                rgb_image = original_image[:, :, :1]
```
russelldj, line 803: It looks like only one channel is being retained. Is this correct?

russelldj (approved):
Looks good to me aside from the one comment. I'd like to test it out but that might be a little while since our cloud compute is offline for the week.

russelldj:
Trying to run this on Windows 11 I'm getting a dependency issue. Installing `SAM2` with pip is upgrading numpy from major version 1 to 2. This breaks the imports for packages compiled against numpy v1. I suspect this would be a problem on unix platforms as well.

tree_detection_framework/detection/detector.py
russelldj: I didn't realize we did that in the other ones. I'd lean toward omitting it, but I think it's fine either way.

tree_detection_framework/detection/detector.py
russelldj: Ok thanks for looking into this

russelldj:
This ran successfully for me after some dependency hacks and fixing the channel slicing. I'm happy to merge this now and deal with dependency issues later.

russelldj:
Since there are still dependency issues with SAM2, I moved it into it's own file.

russelldj:
Thanks @mich889!

PR #63, opened 2024-12-03
Add a post-processing method for the detections 

amrithasp02:
Apply post-processing techniques that include: 
    1. Get a union of polygons which have single tree crowns split as multiple polygons across tiles (happens at the tile edges).
    2. Simplify the edges of polygons by a tolerance value
    3. Remove holes within the polygons that are smaller than a minimum area value


tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -132,3 +136,68 @@ def multi_region_NMS(
...
+
+    # Simplify the polygons by tolerance value and extract only Polygons and MultiPolygons
+    # since `union_detections` can have Point objects as well
+    filtered_geoms = [
+        geom.simplify(tolerance)
+        for geom in list(union_detections.geoms)
+        if isinstance(geom, (Polygon, MultiPolygon))
+    ]
+
+    # To remove small holes within polygons
```
russelldj, line 183: The hole removal part looks really good to me.

russelldj: Maybe it could be its own post-processing step in cases where you don't want to both merge detections and remove holes?

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -132,3 +136,68 @@ def multi_region_NMS(
...
+    # and combines a single polygon that might have been split into multiple. Also removes any overlaps.
+    union_detections = unary_union(all_detections_gdf.geometry)
+
+    # Simplify the polygons by tolerance value and extract only Polygons and MultiPolygons
+    # since `union_detections` can have Point objects as well
+    filtered_geoms = [
+        geom.simplify(tolerance)
+        for geom in list(union_detections.geoms)
+        if isinstance(geom, (Polygon, MultiPolygon))
+    ]
```
russelldj, line 181: If I understand this correctly, each element in `filtered_geoms` is created by merging together a set of detections that overlap each other at all. I think this would correctly work to merge detections at the boundary but I think it might happen in other situations as well. Depending on the detector, we might have partially overlapping detections within a tile as long as the overlap is less than the NMS threshold. I think we'd want to keep these separate. 

Maybe we can brainstorm other ways to suppress only the overlapping detections at the edge? For example, removing some detections within a buffer distance of the boundary or trying a different intersection metric in the IoU computation to suppress detections that are a subset of another one.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -132,3 +136,68 @@ def multi_region_NMS(
...
+            # If the area of the hole is greater than the threshold, include it in the final output
+            if p.area > min_area_threshold:
+                list_interiors.append(interior)
+
+        # Create a new polygon for the same that does not have the smaller holes
+        new_polygon = Polygon(polygon.exterior.coords, holes=list_interiors)
+        new_polygons.append(new_polygon)
+
+    # Create a RegionDetections for the merged and postprocessed detections
+    postprocessed_detections = RegionDetections(
```
russelldj, line 199: One issue I see is this doesn't retain the data, such as score or class, since we did the merging and now there are a different number of polygons. However, the hole removal and simplification parts don't have this issue.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -132,3 +136,68 @@ def multi_region_NMS(
...
+        new_polygons, CRS=crs, input_in_pixels=False
```
russelldj, line 200: I think there's a subtle issue with the CRS. You didn't request the that `all_detections_gdf` be in the user-specified CRS, so you might actually be interpreting the polygons in an incorrect CRS.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -132,3 +136,68 @@ def multi_region_NMS(
...
+    2. Simplify the edges of polygons by `tolerance` value
+    3. Remove holes within the polygons that are smaller than `min_area` value
+    Merges regions into a single RegionDetections.
+
+    Args:
+        detections(RegionDetectionsSet):
+            Detections from multiple regions to postprocess.
+        crs (Optional[pyproj.CRS], optional):
+            What CRS to use. Defaults to None.
+        tolerance (Optional[float], optional):
```
russelldj, line 158: The tolerance and the min_area_threshold are going to be in whatever units the data is. So you could get very different results if it's a meters-based CRS vs. a lat-lon one in degree units. One option would be to drop the option for the user to specify a CRS and instead compute the appropriate meters-based CRS for the data and use that.

russelldj (commented):
Nice work, there's some tricky concepts here. My big conceptual concern here is with the merging operation since the way it is currently implemented it removes the attributes and merges detections I'm not sure should be.

sandbox/detection/geometric_detections.ipynb
russelldj, line 1: It looks like the merging issue I was worried about happens in this file, since there are clumps of trees that I assume were initially individually detected. Visualizing the before and after could be helpful.

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: Makes sense, I'll create a function separately for that

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: I see. I initially tried to get the CRS from the `predictions` `RegionDetectionsSet` object and maintain that till the end. But, when I try `predictions.get_default_CRS()` it returns this:
```
<Bound CRS: PROJCS["unknown",GEOGCS["unknown",DATUM["Unknown b ...>
Name: unknown
Axis Info [cartesian]:
- [east]: Easting (metre)
- [north]: Northing (metre)
Area of Use:
- undefined
Coordinate Operation:
- name: Transformation from unknown to WGS84
- method: Position Vector transformation (geog2D domain)
Datum: Unknown based on GRS80 ellipsoid
- Ellipsoid: GRS 1980
- Prime Meridian: Greenwich
Source CRS: unknown
```

russelldj: Weird, I'm not sure what's up. When you get a chance could you find the minimum reproducable example and open an issue for this? It's strange because it seems like your data is properly georeferenced.

tree_detection_framework/detection/detector.py
```diff
@@ -190,7 +192,13 @@ def get_geospatial_bounds_as_shapely(
     @staticmethod
     def get_CRS_from_batch(batch):
         # Assume that the CRS is the same across all elements in the batch
-        return batch["crs"][0]
+        CRS = batch["crs"][0]
+        CRS = CRS.to_epsg()
```
russelldj, line 196: Maybe to make this slightly more clear, you could combine this line and the one below it into one line with a comment?

tree_detection_framework/detection/detector.py
```diff
@@ -412,8 +420,16 @@ def get_tree_crowns(
             # Thresholding the tile image
             binary_mask = image > threshold
             # Convert the mask to shapely polygons, returned as a MultiPolygon
-            multipolygon_mask = mask_to_shapely(binary_mask)
-            all_multipolygon_masks.append(multipolygon_mask)
+            shapely_polygon_mask = mask_to_shapely(binary_mask)
+            # Iterate through each polygon in the multipolygon
+            for i, polygon in enumerate(shapely_polygon_mask.geoms):
```
russelldj, line 425: Does this `.geoms` attribute work if `shapely_polygon_mask` is a `Polygon` and not `MultiPolygon`?

amrithasp02: It seems like `mask_to_shapely` always returns a `MultiPolygon`. However, I added a check to be safe.

tree_detection_framework/detection/detector.py
```diff
@@ -73,6 +74,7 @@ def predict_as_generator(
             batch_image_bounds = self.get_image_bounds_as_shapely(batch)
             batch_geospatial_bounds = self.get_geospatial_bounds_as_shapely(batch)
             CRS = self.get_CRS_from_batch(batch)
+            CRS = CRS.to_epsg()
```
russelldj, line 77: Is this `to_epsg()` needed anymore? I think since we did that in `get_CRS_from_batch` the returned CRS object should be standardized and can be used directly.

tree_detection_framework/detection/detector.py
```diff
@@ -190,7 +192,12 @@ def get_geospatial_bounds_as_shapely(
     @staticmethod
     def get_CRS_from_batch(batch):
         # Assume that the CRS is the same across all elements in the batch
-        return batch["crs"][0]
+        CRS = batch["crs"][0]
+        # Get the CRS EPSG value and convert it to a pyproj object
```
russelldj, line 196: This is a nitpick, but it might be helpful to add something like "this is to avoid relying on WKT strings which are more likely to be invalid".

tree_detection_framework/postprocessing/postprocessing.py
russelldj: Now that you fixed the issue with the default CRS, what do you think about dropping the user input value? If you need to compute the appropriate projected CRS, the code is [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/367a645235f568106d6a4607bb2f617b77f14f87/tree_detection_framework/utils/geospatial.py#L4).

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -132,3 +136,159 @@ def multi_region_NMS(
...
+        modified_geometries.append(clean_tree_crown)
+
+    # Set this list as the geometry column in the dataframe
+    detections_df.geometry = modified_geometries
+    # Return a new RegionDetections object created using the updated dataframe
+    return RegionDetections(
+        detection_geometries=None,
+        data=detections_df,
+        input_in_pixels=False,
+        CRS=detections.get_CRS(),
```
russelldj, line 210: This could be a problem if the data is in pixels with no transform to geospatial. This is probably an issue we need to think more about in general, so maybe just leave a comment?

russelldj (approved):
This looks good! I'm good with it being merged, there's just a few things around the CRS that could potentially be cleaned up.

PR #69, opened 2024-12-16
Create DataLoader from a folder of raw drone images

amrithasp02:
- Created a new subclass of `Dataset` called `CustomImageDataset`
- `__getitem__` method within this class deals with the main logic to generate tiles from the images
- Added a method `create_image_dataloader` in `preprocessing.py` that handles the dataloader creation steps
- Included a demo notebook to visualize how the images are tiled

I haven't been able to test it with the detectors yet because we cannot save the predictions as a `RegionDetectionsSet` object at this point, as the `bounds` value is missing for this data.

### DIFF
--- a/tree_detection_framework/preprocessing/derived_geodatasets.py
+++ b/tree_detection_framework/preprocessing/derived_geodatasets.py
@@ -1,5 +1,8 @@
 from typing import Any, Optional
+from collections import defaultdict
 
+from pathlib import Path
+from PIL import Image
 import fiona
 import fiona.transform
 import matplotlib.pyplot as plt
@@ -8,7 +11,8 @@ import rasterio
 import shapely.geometry
 import torch
 from shapely.affinity import affine_transform
-from torch.utils.data import DataLoader
+from torch.utils.data import DataLoader, Dataset
+from torchvision import transforms
 from torchgeo.datamodules import GeoDataModule
 from torchgeo.datasets import (
     IntersectionDataset,
@@ -18,6 +22,8 @@ from torchgeo.datasets import (
 )
 from torchgeo.datasets.utils import BoundingBox, array_to_tensor
 from torchgeo.samplers import GridGeoSampler, Units
+
+from tree_detection_framework.constants import PATH_TYPE
 
 
 class CustomRasterDataset(RasterDataset):
@@ -176,6 +182,71 @@ class CustomVectorDataset(VectorDataset):
 
         return sample
 
+class CustomImageDataset(Dataset):
+    def __init__(
+        self,
+        folder_path: PATH_TYPE,
+        chip_size: float,
+        chip_stride: float,
+    ):
+        """
+        Dataset for creating a dataloader froma folder of individual images, with an option to create tiles.
+
+        Args:
+            folder_path (Path): Path to the folder containing image files.
+            chip_size (float): Dimension of each image chip (width, height) in pixels.
+            chip_stride (float): Stride to take while chipping the images (horizontal, vertical) in pixels.
+        """
+        self.folder_path = Path(folder_path)
+        self.chip_size = chip_size
+        self.chip_stride = chip_stride
+        self.image_paths = list(self.folder_path.glob("*.JPG")) + list(self.folder_path.glob("*.png"))
+
+        if not self.image_paths:
+            raise ValueError(f"No image files found in {self.folder_path}")
+
+        self.tile_metadata = self._get_metadata()
+
+    def _get_metadata(self):
+        metadata = []
+        for img_idx, img_path in enumerate(self.image_paths):
+            with Image.open(img_path) as img:
+                img_width, img_height = img.size
+                
+                # Generate tile coordinates
+                for y in range(0, img_height - self.chip_size + 1, self.chip_stride):
+                    for x in range(0, img_width - self.chip_size + 1, self.chip_stride):
+                        metadata.append((img_idx, img_path, x, y))
+        return metadata
+
+    def __len__(self):
+        return len(self.tile_metadata)
+
+    def __getitem__(self, idx):
+        img_idx, img_path, x, y = self.tile_metadata[idx]
+        with Image.open(img_path) as img:
+            img = img.convert("RGB")
+            tile = img.crop((x, y, x + self.chip_size, y + self.chip_size))
+
+            if not isinstance(tile, torch.Tensor):
+                tile = transforms.ToTensor()(tile)
+
+        return {
+            "image": tile,
+            "metadata": {
+                "image_index": img_idx,
+                "source_image": str(img_path),
+                "tile_coords": (x, y),
+            },
+        }
+    
+    @staticmethod
+    def collate_as_defaultdict(batch):
+        # Stack images from batch into a single tensor
+        images = torch.stack([item["image"] for item in batch])
+        # Collect metadata as a list
+        metadata = [item["metadata"] for item in batch]
+        return defaultdict(lambda: None, {"image": images, "metadata": metadata})
 
 class CustomDataModule(GeoDataModule):
     # TODO: Add docstring
--- a/tree_detection_framework/preprocessing/derived_geodatasets.py
+++ b/tree_detection_framework/preprocessing/derived_geodatasets.py
@@ -182,6 +182,7 @@ class CustomVectorDataset(VectorDataset):
 
         return sample
 
+
 class CustomImageDataset(Dataset):
     def __init__(
         self,
@@ -203,7 +204,13 @@ class CustomImageDataset(Dataset):
 
         # Get a list of all image paths
         image_extensions = [".png", ".jpg", ".jpeg", ".bmp", ".tiff"]
-        self.image_paths = sorted([path for path in self.folder_path.glob("*") if path.suffix.lower() in image_extensions])
+        self.image_paths = sorted(
+            [
+                path
+                for path in self.folder_path.glob("*")
+                if path.suffix.lower() in image_extensions
+            ]
+        )
 
         if not self.image_paths:
             raise ValueError(f"No image files found in {self.folder_path}")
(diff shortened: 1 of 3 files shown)


tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+            img = img.convert("RGB")
+
+            # Check if the tile extends beyond the image boundary
+            tile_width = min(self.chip_size, img.width - x)
+            tile_height = min(self.chip_size, img.height - y)
+
+            # If the tile fits within the image, return the cropped image
+            if tile_width == self.chip_size and tile_height == self.chip_size:
+                tile = img.crop((x, y, x + self.chip_size, y + self.chip_size))
+            else:
```
russelldj, line 250: This is smart, but I don't think it will handle chips that are partially over the bounds of the image, which I think is the most common case. What if you take the image crop and then pad it with white if needed?

russelldj: Sorry, I missed the following lines. This is good. You still might be able to do it faster with `np.pad` but I don't know.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
 
+class CustomImageDataset(Dataset):
+    def __init__(
+        self,
+        folder_path: PATH_TYPE,
+        chip_size: int,
+        chip_stride: int,
+    ):
+        """
+        Dataset for creating a dataloader froma folder of individual images, with an option to create tiles.
```
russelldj, line 194: typo: froma

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+            folder_path (Path): Path to the folder containing image files.
+            chip_size (int): Dimension of each image chip (width, height) in pixels.
+            chip_stride (int): Stride to take while chipping the images (horizontal, vertical) in pixels.
+        """
+        self.folder_path = Path(folder_path)
+        self.chip_size = chip_size
+        self.chip_stride = chip_stride
+
+        # Get a list of all image paths
+        image_extensions = [".png", ".jpg", ".jpeg", ".bmp", ".tiff"]
```
russelldj, line 206: This could be a parameter defaulted to this value. What you have now is also completely fine.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+                        tile_idx += 1
+        return metadata
+
+    def __len__(self):
+        return len(self.tile_metadata)
+
+    def __getitem__(self, idx):
+        img_idx, img_path, x, y = self.tile_metadata[idx]
+        with Image.open(img_path) as img:
+            img = img.convert("RGB")
```
russelldj, line 241: Will this always return a given datatype (e.g. uint8) or could it vary? I think tiff can encode various datatypes.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+            if tile_width == self.chip_size and tile_height == self.chip_size:
+                tile = img.crop((x, y, x + self.chip_size, y + self.chip_size))
+            else:
+                # Create a white square tile of shape 'chip_size'
+                tile = Image.new(
+                    "RGB", (self.chip_size, self.chip_size), (255, 255, 255)
+                )
+
+                # Crop the image section and paste onto the white image
+                img_section = img.crop((x, y, x + tile_width, y + tile_height))
```
russelldj, line 257: Do you know if this `PIL.crop` method is better than using normal numpy slicing? It's completely fine to use it, I was just curious if it happened to have additional benefits like not loading the whole image into memory.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+        return {
+            "image": tile,
+            "metadata": {
+                "image_index": img_idx,
+                "source_image": str(img_path),
+                "tile_coords": (x, y),
+            },
+        }
+
+    # TODO: Think of a way to make this a common function for `CustomRasterDataset` and `CustomImageDataset`
```
russelldj, line 273: What would need to change? It seems like it might be pretty easy to just pull it out as a stand-alone function the way it's written.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
         return sample
 
 
+class CustomImageDataset(Dataset):
+    def __init__(
```
russelldj, line 187: In the future we might want to think about having an easy way to not tile the images and instead provide the whole one. This would require that every image in the folder is the same size. Also, we might want to consider the option to downsample the images, but I think that can also wait until later.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -177,6 +183,141 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+        if not self.image_paths:
+            raise ValueError(f"No image files found in {self.folder_path}")
+
+        self.tile_metadata = self._get_metadata()
+
+    def _get_metadata(self):
+        metadata = []
+        for img_path in self.image_paths:
+            tile_idx = 0  # A unique tile index value within this image, resets for every new image
+            with Image.open(img_path) as img:
```
russelldj, line 224: Out of curiosity, does this actually load the image data or just the metadata? I'm just wondering how fast it is.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -172,6 +177,59 @@ def create_dataloader(
...
+    dataset = CustomImageDataset(
+        folder_path=folder_path,
+        chip_size=chip_size,
+        chip_stride=chip_stride,
+    )
+    dataloader = DataLoader(
+        dataset,
+        batch_size=batch_size,
+        shuffle=False,
+        collate_fn=CustomImageDataset.collate_as_defaultdict,
```
russelldj, line 228: Was there a reason you needed to write your own here?

russelldj (approved):
Nice work here, this seems really well thought out. I have some potential simplifications or speed ups, but I'm happy for it to be merged at any point.

russelldj:
If you wanted to make this more consistent with the geospatial dataloader, you could add a `bounds` object that represents the bounds of the cropped sub-region in pixel coordinates relative to the full image. You could also add a `None` CRS. 

It may take some work to get the `RegionDetectionSet` object to work with data that doesn't actually have a CRS. But I think we can figure it out.

If we do standardize it to the point where we can generate `RegionDetections` for each tile, we might have to post-process that result. Specifically, I think we'd want to break it apart by which image generated the predictions, since we'd only want to run across-tile NMS on predictions from the same image.

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: This is a lazy operation. It doesn't load the image until `load()` is explicitly called. So, at this point it just loads metadata like image size

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: It looks like after the `.convert('RGB')` operation it always returns `uint8` datatype. I think even removing this line shouldn't affect because even before this operation, the image is in 'RGB' mode. I just added this line to force it to be in that mode.

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: I think only during the crop operation it loads the image into memory to get the pixel data. I sticked to using `PIL.Image` here since we used it to generate the metadata in the earlier step. We could also using `numpy` here if you'd prefer that.

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: Yeah we can just add it as a `utils` function. I was initially thinking if we wanted to create a parent class and define the `plot` method there. Then subclass `CustomRasterDataset` and `CustomImageDataset` from that so that both can use the `plot` method. But I guess that would complicate things unnecessarily.

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: I'm happy to leave it as-is, especially if there's a chance there's a performance benefit.

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: Either way works. If you think there is additional functionality (beyond this method) that might be shared across the two classes I think a parent class makes sense. Otherwise no strong preference.

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: Perfect

tree_detection_framework/preprocessing/preprocessing.py
amrithasp02: Yes. Basically, the datalaoders we have been working with so far (created by `torchgeo` methods) return the batch as a 4D tensor which first dim being the `batch_size`. However, the default `collate` method for this `torch` DataLoader was a list of tensors.

PR #74, opened 2024-12-20
Accelerate `get_treetops` method for GeometricDetector

amrithasp02:
The optimized algorithm resulted in a speed up of approximately 63%. The GeometricDetector now generates predictions for Emerald Point data in ~35 seconds.

The previous algorithm (renamed as `get_treetops_bruteforce`) which iterates through every pixel, has been retained as an alternative approach for now. Please let me know if it would be more appropriate to remove it if it's no longer necessary.


russelldj (commented):
This looks good. I'll defer to you on whether to leave the `brute_force` implementation. Do you think there are any situations where it would be faster to do the brute force approach or the filtering approach would have too large of a memory footprint? Also, can you re-run the timing experiments once you implement my changes to make sure it doesn't change the results dramatically? It's possible filtering with a circular filter could be slower by a lot, if they have an optimized version for the rectangular case.

russelldj:
Since a lot of the logic is shared between the two methods, you could alternatively refactor `get_treetops` to take a `precompute_local_max/bruteforce` (or similar) keyword argument. And then based on that argument you could either have your `selected_indices` indices be every index above the min height or every index that is both above the min height and a local max.

amrithasp02:
> Since a lot of the logic for checking whether an index is actually a max is shared between the two methods, you could alternatively refactor `get_treetops` to take a `precompute_local_max/bruteforce` (or similar) keyword argument. And then based on that argument you could either have your `selected_indices` indices be every index above the min height or every index that is both above the min height and a local max.

@russelldj 
I decided to include the option to use bruteforce as a default if `filter_shape` argument has value `none`. 



> This looks good. I'll defer to you on whether to leave the `brute_force` implementation. Do you think there are any situations where it would be faster to do the brute force approach or the filtering approach would have too large of a memory footprint? Also, can you re-run the timing experiments once you implement my changes to make sure it doesn't change the results dramatically? It's possible filtering with a circular filter could be slower by a lot, if they have an optimized version for the rectangular case.

I didn't have any specific reasons for keeping the bruteforce method. Anyway it wasn't too much work to include it in the updated function, so I've kept it.

I've also added the results from the timing experiments for `square` `circle` and `none` values in the notebook [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/feature/AP/accelerate-geometric-detector/sandbox/detection/geometric_detections.ipynb) cell 4 onwards. From the results it looks like `square` filter is the fastest. `circle` filter takes around 5 seconds more.

tree_detection_framework/detection/detector.py
```diff
@@ -379,56 +379,96 @@ def calculate_edge_distance(centroid):
...
+            filtered_image = maximum_filter(
+                image, footprint=footprint, mode="constant", cval=0
+            )
 
-                # Create column and row vectors for the neighborhood
-                region_i = np.arange(i_min, i_max)[:, np.newaxis]
-                region_j = np.arange(j_min, j_max)[np.newaxis, :]
+        elif self.filter_shape == "square":
+            # Create a square window using the computed radius. Add 1 to make it odd.
+            window_size = (min_radius_pixels * 2) + 1
```
russelldj, line 417: In some very rare cases, this square filter may suppress detections that the circle or bruteforce method would keep if those detections lie right at the corner of the suppression filter. One option would be to make the filter smaller by a factor of `1/sqrt(2)` so even the corners of the filter are within the suppression radius. Alternatively, maybe including a warning or comment about this would be sufficient.

russelldj (approved):

russelldj:
I like the way you restructured it to deal with the different pre-filtering options.

tree_detection_framework/detection/detector.py
amrithasp02: Thanks for pointing that out!

tree_detection_framework/detection/detector.py
```diff
@@ -379,56 +379,97 @@ def calculate_edge_distance(centroid):
...
+                image, footprint=footprint, mode="constant", cval=0
+            )
 
-                # Create column and row vectors for the neighborhood
-                region_i = np.arange(i_min, i_max)[:, np.newaxis]
-                region_j = np.arange(j_min, j_max)[np.newaxis, :]
+        elif self.filter_shape == "square":
+            # Create a square window using the computed radius. Add 1 to make it odd.
+            # Reduce filter size by 1/sqrt(2) to keep corners within the suppression radius.
+            window_size = int((min_radius_pixels * 2 + 1) / np.sqrt(2))
```
russelldj, line 418: Previously you were ensuring this was an odd integer but this is no longer the case. Does it need to be that?

amrithasp02: I just verified the filtering seems to work fine without adding the 1. So should I leave it as it is? I could also probably remove the 1 from `(min_radius_pixels * 2 + 1)`

russelldj: I think the version you have in the most recent commit is good.

tree_detection_framework/detection/detector.py
```diff
@@ -379,56 +395,97 @@ def calculate_edge_distance(centroid):
...
+            filtered_image = maximum_filter(
+                image, footprint=footprint, mode="constant", cval=0
+            )
 
-                # Calculate the distances to every point within the region
-                distances = np.sqrt((region_i - i) ** 2 + (region_j - j) ** 2)
+        elif self.filter_shape == "square":
+            # Create a square window using the computed radius
+            # Reduce filter size by 1/sqrt(2) to keep corners within the suppression radius
+            window_size = int((min_radius_pixels * 2) / np.sqrt(2))
```
russelldj, line 434: Actually, sorry, one last question. Is it ok that this isn't an integer?

russelldj (commented):
Otherwise everything else looks good

tree_detection_framework/detection/detector.py
amrithasp02: I thought the `int()` type casts it as an integer

russelldj: My bad, I misread the parenthesis and thought the division was happening after the cast. You're correct, sorry.

PR #75, opened 2025-01-02
Adding NMS using IoS and fixes in GeometricDetector

amrithasp02:
- Added a method to perform NMS (IoU + IoS) on tree detections and evaluated qualitatively to confirm it removes the crowns split at tile boudaries.
- Removed normalization of GeometricDetector scores.
- Resolved the occurrence of MultiPolygon crowns in the final prediction output.
- Other minor fixes in the GeometricDetector class.
- Included notebook to demonstrate NMS.


tree_detection_framework/detection/detector.py
```diff
@@ -623,14 +594,14 @@ def predict_batch(self, batch):
         # List to store every image's detections
         batch_detections = []
         batch_detections_data = []
-        for image, bounds in zip(batch["image"], batch["bounds"]):
+        for _, image in enumerate(batch["image"]):
```
russelldj, line 597: Is there a reason you need the enumerate here?

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -295,3 +303,41 @@ def merge_and_postprocess_detections(
...
     return postprocessed_detections
+
+
+def suppress_tile_boundary_with_NMS(
+    predictions: RegionDetectionsSet,
+    threshold: float = 0.5,
+    min_confidence: float = 0.3,
+) -> RegionDetections:
+    """
+    Used as a post-processing step with the `GeometricDetector` class to suppress detections that are split across tiles.
```
russelldj, line 314: Maybe add something brief that this is done by first running NMS with IoU then IoS.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -295,3 +303,41 @@ def merge_and_postprocess_detections(
...
+        predictions,
+        intersection_method="IOU",
+        threshold=threshold,
+        min_confidence=min_confidence,
+    )
+
+    iou_ios_nms = single_region_NMS(
+        iou_nms,
+        intersection_method="IOS",
+        threshold=threshold,
```
russelldj, line 339: I'd consider a different, much higher threshold here.

tree_detection_framework/detection/detector.py
```diff
@@ -554,53 +535,43 @@ def get_tree_crowns(
...
-                for polygon in list(tree_crown.geoms):
-                    # Note that any LineString/Point/MultiPoint objects get ignored. MultiPolygon is rarely found.
-                    if isinstance(polygon, Polygon):
-                        multipolygon.append(polygon)
-                # Finally create an equivalent MultiPolygon object for the GeometryCollection
-                filtered_crowns.append(MultiPolygon(multipolygon))
-
-        # Update the 'tree_crown' column
-        tile_gdf["tree_crown"] = filtered_crowns
+            elif isinstance(tree_crown, (MultiPolygon, GeometryCollection)):
```
russelldj, line 563: Out of curiosity, what was the issue that led to multipolygons being retained initially?

russelldj (approved):
Looks good to me

tree_detection_framework/detection/detector.py
amrithasp02: You're right, it is not needed

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: I tried changing the threshold from 0.5 to 0.9 for the NMS using IoS step. However, the output seems to retain more partially split tree crowns by the boundaries. Is there a reason why you feel a higher threshold would be better?

amrithasp02: Follow-up Q: Should I make the `iou_threshold` and `ios_threshold` as separate parameters passed to the `suppress_tile_boundary_with_NMS` method?

russelldj: I was thinking this would be primarily used in the case where one tile had the full detection and the other had only a subset of it. But I hadn't thought of your case where there is only a partial detection in each tile. Maybe we leave the threshold lower for now and consider revisiting it in the future? For example if we start using larger tiles and we can be more conservative. Or if we notice this is removing detections we'd actually like to retain.

tree_detection_framework/detection/detector.py
amrithasp02: Previously, I was converting all `GeometricCollections` to `MultiPolygons` which is why they were retained in the final detections.

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: Makes sense, thanks for clarifying

tree_detection_framework/detection/detector.py
russelldj: Ok, good catch

PR #76, opened 2025-01-08
Support tree detection in raw drone images

amrithasp02:
- Fix image normalization steps within DeepForest and Detectree2
- Accept `None` type CRS by removing some previous errors being raised by the framework
- Created a new method in `Detector` to handle logic for returning a set of `RegionDetectionsSet` objects
- Notebook with outputs in `tree-detection-framework/sandbox/detection/detect_raw_drone_images.ipynb`


tree_detection_framework/detection/detector.py
```diff
@@ -133,6 +134,54 @@ def predict(
...
+        # computation is defered until the samples are actually requested
+        predictions_generator = self.predict_as_generator(
+            inference_dataloader, **kwargs
+        )
+        # This step is where the computation actually occurs since all samples are requested to
+        # build the list
+        predictions_list = list(predictions_generator)
+
+        # Extract the source image names associated with each tile in the inference dataloader
+        image_filenames = [
```
russelldj, line 164: Smart syntax, I didn't realize you could chain nested iterables together like this.

tree_detection_framework/detection/detector.py
```diff
@@ -199,9 +248,10 @@ def get_CRS_from_batch(batch):
         CRS = batch["crs"][0]
         # Get the CRS EPSG value and convert it to a pyproj object
         # This is to avoid relying on WKT strings which are more likely to be invalid
-        CRS = pyproj.CRS(CRS.to_epsg())
-        if CRS.to_epsg() is None:
-            raise ValueError(f"Invalid CRS. Found CRS = {CRS.to_epsg()}")
+        if CRS is not None:
+            CRS = pyproj.CRS(CRS.to_epsg())
+            if CRS.to_epsg() is None:
+                raise ValueError(f"Invalid CRS. Found CRS = {CRS.to_epsg()}")
```
russelldj, line 254: Won't this always print `CRS = None`? The general logic seems fine, just the error message seems redundant.

tree_detection_framework/detection/detector.py
```diff
@@ -717,8 +767,12 @@ def predict_batch(self, batch):
 
         self.lightningmodule.eval()
         images = batch["image"]
+        # DeepForest requires input image pixel values to be normalized to range 0-1
         with torch.no_grad():
-            outputs = self.lightningmodule(images[:, :3, :, :] / 255)  # .model ??
+            if images.min() >= 0 and images.max() <= 1:
```
russelldj, line 772: I think this is fine for now, but in the future I'd like to make the data loaders more flexible so the user can provide the correct data type/range. That would avoid duplicating this type of logic in each model and allow the pre-processing computation to be performed in a different thread. Does this seem reasonable?

tree_detection_framework/detection/detector.py
```diff
@@ -815,22 +869,35 @@ def call_predict(self, batch):
...
-                    original_image
-                )
-                image = torch.as_tensor(image.astype("float32").transpose(2, 0, 1))
-                image = image.to(self.cfg.MODEL.DEVICE)
-
-                # Create a dict with each image and its properties
-                input = {"image": image, "height": height, "width": width}
-                # Add the dictionary to batch image list
-                inputs.append(input)
+            if batch.shape[1] == 3:  # RGB image
```
russelldj, line 872: Why does this check work?

russelldj: Is this input type not permuted?

tree_detection_framework/detection/region_detections.py
```diff
@@ -498,13 +493,6 @@ def merge(
         Returns:
             gpd.GeoDataFrame: Detections in the requested CRS or in pixel coordinates with a None .crs
         """
-        if not self.all_regions_have_CRS():
-            raise ValueError("Merging requires all sub-regions to have a valid CRS")
-
-        # If the CRS is not set, use that of the first sub-region
```
russelldj, line 504: In the future we should probably revisit this logic together to figure out how we want to deal with the CRS in general. But for now this works.

russelldj (approved):
Great work, this is really exciting.

tree_detection_framework/detection/detector.py
amrithasp02: The dataloaders created using raw drone images have just the RGB channels whereas those created using the raster orthomosaic data have 4 channels (alpha channel?). The preprocessing steps were different for both so I put this check here.

tree_detection_framework/detection/detector.py
amrithasp02: Sounds good

PR #78, opened 2025-01-08
Add updated poetry lock from updated polygone-NMS

russelldj:
This appears to fix the dependency conflicts between `SAM2` and `polygone-NMS`. The NMS requirements were previously pinned to a specific numpy v1 version [here](https://github.com/open-forest-observatory/PolyGoneNMS/blob/main/requirements.txt). I updated that to v2 and re-solved the lock file for TDF with these new sub-requirements.


amrithasp02 (approved):
Nice!

PR #80, opened 2025-01-14
Removing detections outside image bounds

amrithasp02:
A post-processing step for predictions obtained from dataloader created with raw drone images


tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -345,3 +346,71 @@ def suppress_tile_boundary_with_NMS(
...
+            Each elemet is a RegionDetectionsSet derived from a specific drone image.
+            Length is the number of raw drone images given to  the dataloader.
+        true_bounds (List[bounding_box]):
+            Each element is a `bounding_box` object derived from the dataloader.
+            Length: number of regions in a set * number of RegionDetectionSet objects
+
+    Returns:
+        List of RegiondetectionSet objects with out-of-bounds predictions filtered out.
+    """
+    # Find the number of regions in every set. Assume every set has same number of regions.
```
russelldj, line 369: This is a reasonable assumption for now. However, there are various drone models included in the OFO catalog and some have different sized images.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -345,3 +346,71 @@ def suppress_tile_boundary_with_NMS(
...
+            # This is essentially the dimensions of the image
+            region_set_polygon = box(minx, miny, maxx, maxy)
+
+            # Remove detections that extend beyond the image bounds
+            indices_within_bounds = rd_gdf[
+                rd_gdf["geometry"].apply(lambda poly: poly.within(region_set_polygon))
+            ].index
+
+            # Subset the RegionDetections object keeping only the valid indices calculated before
+            filtered_rd = rd.subset_detections(list(indices_within_bounds))
```
russelldj, line 406: This can allow a detection that is partially over the bounds of the image, correct? I could see the motivation for not only removing detections that do not overlap with the image but also cropping each detection to fully be contained within the image.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -345,3 +346,71 @@ def suppress_tile_boundary_with_NMS(
...
+    list_of_filtered_region_sets = []  # To save the final result
+
+    for rds in region_detection_sets:
+
+        # To save filtered regions in a particular set
+        list_of_filtered_regions = []
+
+        # Subset true_bounds for the current region set
+        region_true_bounds_set = true_bounds[
+            region_idx : region_idx + num_of_regions_in_a_set
```
russelldj, line 381: To avoid the assumption that each region detection set has the same number of regions, could you instead compute the number of detections in each `rds`?

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -231,7 +231,15 @@ def __getitem__(self, idx):
             "metadata": {
                 "image_index": img_idx,
                 "source_image": str(img_path),
+                # True bounds includes bounding box values for image data within a tile
+                "true_bounds": bounding_box(
```
russelldj, line 235: The idea of having `true_bounds` is smart and I can see that being helpful for a variety of tasks. One alternative way to approach it that might be a little more narrowly suited to this particular task is record the image size for each same in the data loader. Then you would not need to compute the min and max bounds of the set of `true_bounds` for each RDS.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -345,3 +346,71 @@ def suppress_tile_boundary_with_NMS(
...
+        ]
+
+        for idx in range(num_of_regions_in_a_set):
+
+            # Get RegionsDetections object from the set
+            rd = rds.get_region_detections(idx)
+            rd_gdf = rd.get_data_frame()
+
+            # Calculate the overall bounding box for the current true bounds subset
+            minx = min(bounds.minx for bounds in region_true_bounds_set)
```
russelldj, line 391: I think these can be calculated outside of the loop. Also, `minx` and `miny` should always be 0, but it's fine to compute them just in case.

russelldj: Also, since you're filtering on a per-`RegionDetection` level, you could consider filtering each one not by the full bounds of the image but just by the `true_bounds` of that one tile. Even if a prediction is within the extent of the image, I think it's reasonable to drop it if it's outside of the chip used to generate it. In practice I can't see a situation where this happens though. 

Yet another consideration, we could make a method of the `RegionDetection{Set}` classes that drops all detections outside of a region. The logic would stay similar, it could just be used in other applications as well. Not required for this PR, just something to think about.

russelldj (commented):
Looks good to me. I think their might be a few ways to make this a little cleaner, but I'd like to hear your thoughts on it. If you think this is the best way to go, I'm happy to merge as-is.

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: This method removes the entire detection even if a portion of the bounding box lies outside `region_set_polygon` (dimensions of the entire image). This usually happens in cases where the detection box extends to the white padded region. Do you feel it would be better to just crop the bounding box instead?

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: That would actually make things a lot easier for this task. I think I chose to add `true_bounds` just to retain additional metadata at the tile-level if any other tasks in the future end up needing it. Do you think it makes sense to keep it as it is or should I switch to recording the image size?

tree_detection_framework/postprocessing/postprocessing.py
amrithasp02: Should I create an issue for this?

tree_detection_framework/preprocessing/derived_geodatasets.py
russelldj: Maybe just add both? Otherwise probably just switch to the image size and add the `true_bounds` back if needed at a later date.

tree_detection_framework/postprocessing/postprocessing.py
russelldj: Sure, thank you

tree_detection_framework/postprocessing/postprocessing.py
russelldj: Ok, I misinterpreted the logic, that makes sense. If it's fairly easy to implement I'd prefer the crop-to-bounds approach just so we can get partial detections at the tile bounds, even if they have to spill over a little bit. But if it's hard it's not important for now.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -345,3 +346,68 @@ def suppress_tile_boundary_with_NMS(
...
+        List of RegiondetectionSet objects with out-of-bounds predictions filtered out.
+    """
+
+    region_idx = 0  # To index elements in true_bounds
+    list_of_filtered_region_sets = []
+
+    for rds in region_detection_sets:
+
+        # Find the number of regions in every set
+        num_of_regions_in_a_set = len(rds.get_data_frame())
```
russelldj, line 376: We should probably make a convenience method for getting the number of regions in the detection set in the future.

russelldj (approved):
I like the new changes, I'm good to merge whenever

PR #84, opened 2025-01-30
Update `polygone_nms` dependency on Poetry

amrithasp02:
Switched to using the ofo-fork version of `PolyGoneNMS` published on `pypi` . Regenerated `poetry.lock`.

Unrelated, but while testing I also noticed a small bug while getting detections from `detectree2` using `mask_to_shapely` with the new `cv2` backend. So I switched the default for that method to `contourpy` and now am passing `backend` as an argument to `GeometricDetector` to support the `cv2` backend as well.


tree_detection_framework/detection/detector.py
```diff
@@ -348,6 +348,7 @@ def __init__(
         threshold_factor: float = 0.3,
         confidence_factor: str = "height",
         filter_shape: str = "circle",
+        backend: str = "cv2",
```
russelldj, line 351: Maybe update this to something more descriptive like `contour_backend`? Though the documentation alone may be sufficient.

sandbox/detection/detect_raw_drone_images.ipynb
russelldj, line 1: As we continue developing we'll probably want to be a bit more careful about checking in notebooks, at least when they have outputs included. Just to avoid the repo history getting too large or having distracting changes.

russelldj (approved):

sandbox/detection/detect_raw_drone_images.ipynb
amrithasp02: I agree. I deleted few of output images.

I also noticed some methods in the notebook having outdated parameters. Corrected those in this, but I'm sure some of the other notebooks may also have them

PR #86, opened 2025-04-04
Add NEON & detectree2 benchmark and evaluation methods

amrithasp02:
This PR focuses on integrating @janehwu's code with TDF.

Main modifications include:

1. Adding methods to calculate IoU, Hungarian matching and precision-recall metrics for `shapely` objects in `tree_detection_framework/evaluation/evaluate.py`. 
2. `get_neon_detections()` in `utils/benchmarking.py` to load NEON images and annotations, get predictions using the selected detector(s).
3. `evaluate_detections()` to compute and report P R values.
4. Outputs and execution saved to `tree-detection-framework/sandbox/evaluation/neon_benchmark.ipynb`

NOTE:

- I have only tested with DeepForest and Detectree2. Will evaluate with SAM2 once this approach is finalized.
- The results currently seen in `neon_benchmark.ipynb` vary from what is in the paper. I suppose this is because NMS is not done on the detections (TODO).


tree_detection_framework/evaluation/evaluate.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    if not ground_truth_boxes or not predicted_boxes:
+        return 0.0  # Return 0 if either list is empty
+
+    num_gt = len(ground_truth_boxes)
+    num_pred = len(predicted_boxes)
+
+    # Create IoU cost matrix (negative because Hungarian minimizes cost)
+    cost_matrix = np.zeros((num_gt, num_pred))
+    for i, gt in enumerate(ground_truth_boxes):
```
russelldj, line 40: This approach is correct but I think you may be able to optimize the runtime if you can make it parallel. You can use [geopandas.overlay](https://geopandas.org/en/stable/docs/reference/api/geopandas.overlay.html) to compute the intersection of each polygon in `GeoDataFrame`  with the polygons in another gdf. There's a couple of subtle steps for how to actually use this to get your IoU matrix. Do you want to try to work through it on your own?

tree_detection_framework/evaluation/evaluate.py
```diff
@@ -0,0 +1,67 @@
...
+        num_gt (int): Number of ground truth boxes.
+        num_pd (int): Number of predicted boxes.
+        threshold (float): IoU threshold for considering a match.
+    Returns:
+        tuple: Precision and recall values.
+    """
+    true_positives = (np.array(ious) > threshold).astype(np.uint8)
+    recall = np.sum(true_positives) / num_gt
+    precision = np.sum(true_positives) / num_pd
+    return round(precision, 3), round(recall, 3)
```
russelldj, line 67: I'd avoid rounding intermediate results and wait until this is displayed to do so.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -151,34 +151,38 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
+                [
+                    path
+                    for path in self.folder_path.glob("*")
+                    if path.suffix.lower() in image_extensions
+                ]
+            )
 
-        if not self.image_paths:
-            raise ValueError(f"No image files found in {self.folder_path}")
+            if not self.image_paths:
```
russelldj, line 182: I usually avoid using `not` for checking zero-length lists, maybe use a explicit length check instead. Just preference though.

tree_detection_framework/utils/benchmarking.py
```diff
@@ -0,0 +1,120 @@
...
+            bndbox = obj.find("bndbox")
+            if bndbox is not None:
+                xmin = int(bndbox.find("xmin").text)
+                ymin = int(bndbox.find("ymin").text)
+                xmax = int(bndbox.find("xmax").text)
+                ymax = int(bndbox.find("ymax").text)
+                gt_boxes.append(box(xmin, ymin, xmax, ymax))
+
+        # Add the ground truth boxes to the mappings
+        mappings[str(path)] = {"gt": gt_boxes}
```
russelldj, line 59: I think it could be clearer if the code for extracting the groundtruth annotations was in its own function. What do you think?

russelldj: Nice work parsing this btw

tree_detection_framework/utils/benchmarking.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Create dataloader setting image size as 420x420. NEON dataset has a standard size of 400x400.
```
russelldj, line 61: Nitpick: why does the stride need to be bigger than the standard NEON size?

tree_detection_framework/utils/benchmarking.py
```diff
@@ -0,0 +1,120 @@
...
+        region_detection_sets, filenames, _ = detector.predict_raw_drone_images(
+            dataloader
+        )
+
+        # Add predictions to the mappings so that it looks like:
+        # {"image_path_1": {"gt": gt_boxes, "detector_name_1": [boxes], ...},
+        #  "image_path_2": {"gt": gt_boxes, "detector_name_1": [boxes], ...}, ...}
+        for filename, rds in zip(filenames, region_detection_sets):
+            gdf = rds.get_data_frame()[0]
+            if name == "deepforest":
```
russelldj, line 81: Maybe add a comment to explain you're converting the output of different detectors into a standard bbox here?

tree_detection_framework/utils/benchmarking.py
```diff
@@ -0,0 +1,120 @@
...
+                ymin = int(bndbox.find("ymin").text)
+                xmax = int(bndbox.find("xmax").text)
+                ymax = int(bndbox.find("ymax").text)
+                gt_boxes.append(box(xmin, ymin, xmax, ymax))
+
+        # Add the ground truth boxes to the mappings
+        mappings[str(path)] = {"gt": gt_boxes}
+
+    # Create dataloader setting image size as 420x420. NEON dataset has a standard size of 400x400.
+    dataloader = create_image_dataloader(
```
russelldj, line 62: Longer term, we could consider whether this dataloader could contain the ground truth information.

tree_detection_framework/evaluation/evaluate.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Solve optimal assignment using the Hungarian algorithm
+    gt_indices, pred_indices = linear_sum_assignment(cost_matrix)
+
+    # Extract IoUs of matched pairs
+    matched_ious = [-cost_matrix[i, j] for i, j in zip(gt_indices, pred_indices)]
+    return matched_ious
+
+
+def compute_precision_recall(
```
russelldj, line 52: Deep forest also computes an average IoU for matched detections. It might be worth including that here too.

tree_detection_framework/utils/benchmarking.py
```diff
@@ -0,0 +1,120 @@
...
+        all_predictions_R = []
+        for img in img_paths:
+            gt_boxes = detections_dict[img]["gt"]
+            pred_boxes = detections_dict[img][detector]
+            iou_output = compute_matched_ious(gt_boxes, pred_boxes)
+            P, R = compute_precision_recall(iou_output, len(gt_boxes), len(pred_boxes))
+            all_predictions_P.append(P)
+            all_predictions_R.append(R)
+
+        P = np.mean(all_predictions_P)
```
russelldj, line 117: This seems like we're macro-averaging across images, since the precision and recall of each image is counted equally, no matter how many predictions/gts there are from that image. This is consistent with @janehwu 's code (as far as I understood it) but we may want to check that it's consistent with other implementations. And potentially allow the user to choose between macro-averaging images and weighting them by the number of true trees.

sandbox/evaluation/neon_benchmark.ipynb
russelldj, line 1: I can't comment line-by-line in browser. This looks really good, I just might make the path to the dataset a single constant at the top of the script.

README.md
```diff
@@ -86,3 +86,15 @@ up-to-date examples in the `examples` folder. We also have work-in-progress or o
 `sandbox`, which still may provide some insight but is not guaranteed to be current or generalizable.
 Finally, the `tree_detection_framework/entrypoints` folder has command line scripts that can be run
 to complete tasks.
+
+## Evaluation and benchmark with NEON
+Download the NEON dataset files
+```
```
russelldj, line 92: Maybe suggest that the user puts both the annotations and RGB in a new folder?

russelldj (changes requested):
This is really impressive. I was able to run it on my end and everything looked good.

tree_detection_framework/utils/benchmarking.py
amrithasp02: Keeping it as 400x400 gives this error:
`GEOSException: IllegalArgumentException: Points of LinearRing do not form a closed linestring`

So I decided to have the extra 20x20 as a white space buffer

tree_detection_framework/utils/benchmarking.py
amrithasp02: I agree. I guess that would involve having the `CustomVectorDataset` object created within the method

tree_detection_framework/utils/benchmarking.py
amrithasp02: Yes, I've separated it from the function

tree_detection_framework/utils/benchmarking.py
russelldj: I'm not sure what would cause this exception, do you know? I worry that it could come up in other future situations. Maybe open an issue for it, even if we have to keep the padding here for the time being?

tree_detection_framework/utils/benchmarking.py
russelldj: Smart, I forgot about that class. I think that would be a clean way to handle it.

tree_detection_framework/utils/benchmarking.py
amrithasp02: There's an explanation [here](https://gis.stackexchange.com/a/97360). I think few tiles extend a little beyond 400x400 so it makes a new tile for the remaining small portion and that is causing this. I'm not 100% sure though. I'll open an issue for it.

amrithasp02:
With the latest commits, I have included benchmark capabilities for Detectree2 site datasets. Now, SAM2 can also be evaluated. See `sandbox/evaluation/sam2_dtree2_benchmark.ipynb` and `sandbox/evaluation/sam2_dtree2_benchmark.ipynb` for SAM2 evaluation.

tree_detection_framework/preprocessing/derived_geodatasets.py
```diff
@@ -151,40 +151,63 @@ def __getitem__(self, query: BoundingBox) -> dict[str, Any]:
...
         self.tile_metadata = self._get_metadata()
 
     def _get_metadata(self):
         metadata = []
         for img_path in self.image_paths:
+            # Ensure the label path corresponds to the same file name as the image
+            label_path = None
+            if self.labels_paths:
+                # Match the label file by replacing the image extension with `.geojson`
+                expected_label_name = img_path.stem + ".geojson"
```
russelldj, line 199: I'm worried there could be a variety of ways that the label information gets specified that wouldn't be compatible with this, such as other extensions (e.g. csv) or a single file for all annotations. Maybe we could make a separate class for the annotation information so this one doesn't get too complex?

tree_detection_framework/utils/benchmarking.py
```diff
@@ -0,0 +1,210 @@
...
+    """Create a Detectree2 dataloader from one or more image/annotation directories."""
+
+    # Extract the list of paths to image and annotation directories
+    images_dirs = [
+        Path(p) for p in (images_dir if isinstance(images_dir, list) else [images_dir])
+    ]
+    annotations_dirs = [
+        Path(p)
+        for p in (
+            annotations_dir if isinstance(annotations_dir, list) else [annotations_dir]
```
russelldj, line 103: clever pattern

russelldj (approved):
This is a lot of really cool work. Longer term, we may want to discuss how to make this type of evaluation generalize well to arbitrary datasets. But I know we're on a short timeline for this work, so I'd just merge it in the current form.

tree_detection_framework/detection/detector.py
```diff
@@ -852,6 +860,10 @@ class Detectree2Detector(LightningDetector):
     def __init__(self, module):
         # TODO: Add lightning module implementation
         # Note: For now, `module` only references to `cfg`
+        if DETECTRON2_AVAILABLE is False:
```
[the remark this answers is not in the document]
amrithasp02: I am raising an ImportError in the beginning [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/feature/AP/support-benchmarking/tree_detection_framework/detection/SAM2_detector.py#L17) if the user runs it in an env without SAM2 installed. So I thought it's not required to add another error in the `__init__` method. 

In case of `Detectree2Detector` I had to have the error in the `__init__` because even the SAM2 detector uses the `Detector` class defined in the same script. So, I'm just have a warning at the top of the script and not an error.

tree_detection_framework/preprocessing/derived_geodatasets.py
amrithasp02: I think that's a great idea. When we create an evaluation function that generalizes to arbitrary datasets, I'll include that. At least for now, I think it's safe to assume detectree2 annotation files are always .geojson files since their tiling code saves the dataset like that.

tree_detection_framework/evaluation/evaluate.py
amrithasp02: Sorry, I somehow entirely missed this comment. Do you want me to make a new PR and add it in?

PR #91, opened 2025-04-17
Take DT2 bbox directly from model

russelldj:
Update detectree2 bounding boxes to be computed directly from the model output.

### DIFF
--- a/tree_detection_framework/detection/detector.py
+++ b/tree_detection_framework/detection/detector.py
@@ -973,17 +973,29 @@ class Detectree2Detector(LightningDetector):
             pred_masks = instances.pred_masks.numpy()
             # Convert each mask to a shapely multipolygon
             shapely_objects = [mask_to_shapely(pred_mask) for pred_mask in pred_masks]
-            all_geometries.append(shapely_objects)
 
-            # Compute axis-aligned minimum area bounding box as Polygon objects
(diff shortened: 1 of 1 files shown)


tree_detection_framework/detection/detector.py
```diff
@@ -973,17 +973,29 @@ def predict_batch(self, batch):
             pred_masks = instances.pred_masks.numpy()
             # Convert each mask to a shapely multipolygon
             shapely_objects = [mask_to_shapely(pred_mask) for pred_mask in pred_masks]
-            all_geometries.append(shapely_objects)
 
-            # Compute axis-aligned minimum area bounding box as Polygon objects
-            bounding_boxes = [box(*polygon.bounds) for polygon in shapely_objects]
+            # Get the model-predicted boxes, which may be larger than the axis-aligned bounding box
```
amrithasp02, line 977: Just out of curiosity, do you know why the model-predicted boxes are larger?

amrithasp02 (approved):

tree_detection_framework/detection/detector.py
russelldj: My understanding is the mask is generated for this initial bounding box. So if the mask doesn't extend to the edges of the region, the minimum bounding box from the mask polygon won't quite be to the edges.

PR #92, opened 2025-04-17
Use bounding boxes for detectree2 NMS

russelldj:
This updates the geometry to the `bbox` column before running NMS.


tree_detection_framework/utils/benchmarking.py
```diff
@@ -81,8 +81,8 @@ def get_neon_dataloader(image_paths: List[str]) -> DataLoader:
     # Create dataloader setting image size as 420x420. NEON dataset has a standard size of 400x400.
```
amrithasp02, line 81: nitpick: could you also update the size in the comment please?

amrithasp02 (approved):
Looks great!

PR #93, opened 2025-04-17
Update benchmarking results in evaluation notebooks

amrithasp02:


russelldj (approved):
Looks good to me. The only thing I wanted to discuss was your (offline) comment about different parameterization of SAM2 with better performance.

tree_detection_framework/detection/detector.py
```diff
@@ -910,32 +910,24 @@ def call_predict(self, batch):
...
                         original_image = original_image * 255
-
-                    # Create a dict with each image and its properties
-                    input = {"image": original_image, "height": height, "width": width}
-
-                    # Add the dictionary to batch image list
-                    inputs.append(input)
+                    original_image = original_image.permute(1, 2, 0).byte().numpy()
             else:
                 for original_image in batch:
```
russelldj, line 918: Nitpick: this might be a little cleaner if there was only one place where you looped over the images in the batch, and then all the conditional logic was handled inside the loop.

tree_detection_framework/detection/detector.py
```diff
@@ -910,32 +910,24 @@ def call_predict(self, batch):
             if batch.shape[1] == 3:  # RGB image
                 for original_image in batch:
                     height, width = original_image.shape[1], original_image.shape[2]
-
                     # Convert image pixel values to 0-255 range
                     if original_image.min() >= 0 and original_image.max() <= 1:
```
russelldj, line 914: I know we're probably not at this point yet, but I'd like to avoid this type of preprocessing at the detector level in the future if possible and have all detectors somehow get the appropriate type of data from the data loader.

tree_detection_framework/detection/detector.py
```diff
@@ -910,32 +910,24 @@ def call_predict(self, batch):
...
-                    inputs.append(input)
+
+            # Apply the augmentation transform to the original image
+            image = self.aug.get_transform(original_image).apply_image(original_image)
+            image = torch.as_tensor(image.astype("float32").transpose(2, 0, 1))
+            image = image.to(self.cfg.MODEL.DEVICE)
+            # Create a dict with each image and its properties
+            input = {"image": image, "height": height, "width": width}
+            # Add the dictionary to batch image list
+            inputs.append(input)
```
russelldj, line 930: It seems like this logic should be within the loop over the batch. Otherwise, `inputs` will only be length one, right?

russelldj (changes requested):
Really nice find

russelldj:
We should probably include somewhere instructions for downloading the Detectree2 `230103_randresize_full.pth` model.

PR #106, opened 2025-05-30
Add option to return only tree top location in geometric detector

russelldj:
Some applications may use just the tree tops and not the crown. It would be possible to record the location of the tree tops as an additional attribute, but these attributes would not be transformed to difference CRS in the same way as the geometry. Additionally, crown delineation takes more time than treetop detection. 

There may be a cleaner way to handle this in the future.


amrithasp02 (approved):
Looks good! I guess in cases where just the treetops are returned, there wouldn't be any need for post-processing? Maybe just filtering duplicate treetop detections in tile edges if there are overlaps? I'm not too sure

Issue #107, opened 2025-06-06
Consider supporting multiple geometry columns


russelldj:
Currently we only have the concept of one geometry column per detection. This parallels the default model for geopandas. However, there are cases where there may be multiple relevant geometric attributes that we want to record for one object. For example, the tree tops and tree crowns for the geometric detector or the predicted bounding box and polygon segmentation mask for a learning-based approach.

The main issue is we'd need to transform these additional geometries in a consistent manner to our `active` geometry. At first glance, I see a couple places where we'd need to make modifications. One is [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/661fded8c3cd3ac7585ec9bb5d4606778bca7494/tree_detection_framework/detection/region_detections.py#L209), where the geometry is transformed from pixel to geospatial coordinates. Another is [here ](https://github.com/open-forest-observatory/tree-detection-framework/blob/661fded8c3cd3ac7585ec9bb5d4606778bca7494/tree_detection_framework/detection/region_detections.py#L271), where the data can be returned in an arbitrary CRS or in pixel coordinates.

`GeoDataFrame` objects can have multiple `GeoSeries` objects, but only the `active` one is transformed by a `to_crs(...)` call. In place of that, we'd need to loop through all the relevant columns and do a `to_crs(...)` call on each individually. The June 2nd, 2021 post [here](https://github.com/geopandas/geopandas/issues/1193) shows a couple examples of how to do that. We could either have the user provide a list of columns to transform or just assume that every geometry-typed column should be transformed, as in the second example. I can't think of a situation off the top of my head where there would be geometry-typed data and we wouldn't want to transform it, but maybe one exists.

PR #108, opened 2025-06-06
Support multiple geometry columns

russelldj:
This partially addresses #107 by supporting transforming the CRS of multiple geometry-typed columns. Which columns are transformed is controlled by a new `geometry_columns` attribute in the constructor. If left default (`None`), all geometry-typed columns will be transformed. The `get_data_frame` method is the only place where this is implemented, so it's important that future methods use this accessor, rather than using the `.detections` attribute directly.

Note that this does not handle transforming multiple columns in pixel coordinates into geospatial coordinates in the constructor, which is what I initially proposed in #107. I think it's best to wait to see how we do the changes discussed recently and whether there's a situation where we would like to provide multiple pixel-coordinate geometry columns at once. Even beyond that discussion, it may be worth rethinking the constructor of this class and whether some of the complexity can be moved to another class, like the Detector.


amrithasp02 (approved):
Looks great!

PR #111, opened 2025-06-10
Apply and arbitrary function to the underlying `.detections` member variable

### LINKED ISSUE #110: Support arbitrary operations on `RegionDetection{Set}` underlying `detections` member variable
Especially when thinking about point-based NMS, I'm finding multiple situations where I'd like to apply a simple operation to the underlying `detections` `GeoDataFrame`. These include buffering, clipping, and finding the centroid of each detection. Ideally, we shouldn't be directly modifying this member variable, but writing a new method for each of these operations becomes tedious. I've currently been following the pattern [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/835cca249420140377dcdd57b73c1e120d17cec9/tree_detection_framework/detection/region_detections.py#L247) of deep-copying the underlying object and then applying the function to the underlying detections and returning the updated `RegionDetections` object. I think this could be easily generalized to take an arbitrary function.

russelldj:
This closes #110 , which provides the context for the issue.

### DIFF
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -519,6 +519,29 @@ class RegionDetectionsSet:
         """
         self.region_detections = region_detections
 
+    def apply_function_to_detections(
+        self,
+        func: Callable[[gpd.GeoDataFrame], gpd.GeoDataFrame],
+        inplace: bool = False,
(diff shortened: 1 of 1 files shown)


amrithasp02 (approved):
Nice!

Issue #109, opened 2025-06-06
Make the `GeometricDetector` two sequential detectors


russelldj:
The `GeometricDetector` works by first detecting tree tops using a local maximum filter and then segments the crowns. These steps occur on a per-tile basis, first finding the [treetops](https://github.com/open-forest-observatory/tree-detection-framework/blob/661fded8c3cd3ac7585ec9bb5d4606778bca7494/tree_detection_framework/detection/detector.py#L740) then segmenting the [crowns](https://github.com/open-forest-observatory/tree-detection-framework/blob/661fded8c3cd3ac7585ec9bb5d4606778bca7494/tree_detection_framework/detection/detector.py#L744). We'd like to move this to two sequential detectors, so we can easily detect only tree tops or see the crown segmentation with results from another source. This will require

PR #112, opened 2025-06-10
Run NMS on points

russelldj:
This implements NMS for point data by suppressing points within a threshold distance of other points. It uses the following steps
- Buffer each point to a circle with radius half the threshold distance
- Run `PolyGoneNMS` with a tiny overlap threshold so almost any will count
- Compute the centroids of the NMS-suppressed circles and return this data.

I considered some other options. `PolyGoneNMS` could be easily modified to add a different method for converting the "similarity score" (instead of IoU, IoS, etc) between two geometric objects, including the distance between them. However, a major part of the code is partitioning the problem into non-overlapping groups of objects that are processed independently. Since point data will never overlap unless the points are identical, no suppression would be performed except for overlapping points, even with a distance-based overlap metric. It would also be possible to write our own NMS function, since even a brute-force computation of N^2 Euclidean distances would likely be tractable in most cases. However, this seems like a lot of code to duplicate.

This addresses one of the two methods for doing point-based suppression from #109 .

### DIFF
--- a/tree_detection_framework/postprocessing/postprocessing.py
+++ b/tree_detection_framework/postprocessing/postprocessing.py
@@ -151,8 +151,18 @@ def multi_region_NMS(
     return NMS_suppressed_merged_detections
 
 
-def update_gdf_to_centroid(gdf: GeoDataFrame):
+def update_gdf_to_centroid(gdf: GeoDataFrame) -> GeoDataFrame:
+    """Return the geodataframe with the geometry updated to be the centroid Point
+
+    Args:
+        gdf (GeoDataFrame): A dataframe to convert
+
+    Returns:
+        GeoDataFrame: the input geodataframe with the geometry updated to be the centroid Point
+    """
+    # Copy the data to avoid updating the orignal
     copied_gdf = gdf.copy()
+    # Set the geometry column to the centroid
     copied_gdf.geometry = gdf.centroid
 
     return copied_gdf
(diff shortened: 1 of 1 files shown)


tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -151,6 +151,74 @@ def multi_region_NMS(
...
+    # Define a function to buffer each point out by half the suppression distance
+    def buffer_out(gdf: GeoDataFrame):
+        copied_gdf = gdf.copy()
+        # Buffer by half the distance because both objects will be buffered
+        copied_gdf.geometry = copied_gdf.buffer(threshold_distance / 2)
+        return copied_gdf
+
+    # Buffer the points out to form circles of the desired radius
+    # TODO Consider error checking that this is actually point-typed data
+    buffered_points = detections.apply_function_to_detections(buffer_out, inplace=False)
```
amrithasp02, line 200: Small clarification: since `buffer_out()` creates a copy of the `gdf` from the `RegionDetections`, if `inplace` was set to `True`, will `buffer_out` still modify the same  `detections` object?

amrithasp02 (approved):

tree_detection_framework/postprocessing/postprocessing.py
russelldj: Good question, this is subtle. From my understanding, if `inplace` was `True`, this would overwrite the `detections` attribute of the input `RegionDetections` rather than creating a copy of the `RegionDetections` and modifying that. However, as you pointed out, `buffer_out` creates a copy. So the initial `detections` object from the input `RegionDetection` would still be unchanged if it were still in scope. However, it gets overwritten [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/764163069412f7a8e5904f09ca2c85886effba55/tree_detection_framework/detection/region_detections.py#L266), so the only reason it would remain in scope is if someone had previously added another reference to the object, such as by doing `initial_detections = un_buffered_RDS.detections` prior to calling `apply_functions_to_detections`. Does this get at your question?

I'm relatively confident on my understanding of Python scoping/garbage collection, but I could have missed something and be off here.

amrithasp02: Yes that makes sense, thanks!

PR #114, opened 2025-06-13
Add function for removing edge detections

russelldj:
This addresses the second bullet of #109, by allowing the user to remove detections at the bounds of a tile so each location is only predicted by the center of one tile. To follow that logic, the suppression distance should be set to `(chip size - chip stride) * output resolution / 2`, based on the parameters used to create the dataloader.

### DIFF
--- a/tree_detection_framework/postprocessing/postprocessing.py
+++ b/tree_detection_framework/postprocessing/postprocessing.py
@@ -1,6 +1,7 @@
 import logging
 from typing import List, Optional, Union
 
+import geopandas as gpd
 import numpy as np
 from geopandas import GeoDataFrame
 from polygone_nms import nms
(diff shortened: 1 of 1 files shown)


amrithasp02 (approved):
Looks good to me!

Now that we have this, do you think [remove_out_of_bounds_detections()](https://github.com/open-forest-observatory/tree-detection-framework/blob/67bcee63dd68f1f1bf70881b1a2334659f7bad22/tree_detection_framework/postprocessing/postprocessing.py#L419) can be simplified to call this method internally for every `RDS` in the list? I had originally created it to postprocess the outputs from `predict_raw_drone_images()` which have the detections as `List[RegionDetectionsSet]` type.

PR #121, opened 2025-06-17
Add UID column for the detections

amrithasp02:
This PR handles including the "unique_ID" column to every `RegionDetectionsSet` object. The main reason for including this is to maintain a mapping of treetop and treecrowns after we get the outputs from `GeometricTreeCrownDetector`


tree_detection_framework/detection/detector.py
```diff
@@ -757,13 +758,35 @@ def predict_batch(self, batch):
...
-                    "height": detected_crowns_gdf["treetop_height"].tolist(),
+
+            data = {
+                "score": confidence_scores,
+                "height": detected_crowns_gdf["treetop_height"].tolist(),
+            }
+
+            # Include the tree top's UID if the input data had it.
+            # This helps map the tree crown to its tree top
+            if "unique_ID" in attribute:
```
russelldj, line 769: This is a clever approach to handle the subsetting. Completely optional suggestion: since you're already tracking additional attributes through the subsetting process in `get_tree_crowns`, it might be simpler to handle it there. You could pass the UID to that method and then add it to the `tile_gdf` when you construct it [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/33c4a4bb86f8575daee727627ea0b3ac58ce728b/tree_detection_framework/detection/detector.py#L644).

russelldj (approved):
Nice work. I have two optional suggestions. Up to you on whether to take them, I know it would be a substantial rewrite.

tree_detection_framework/detection/detector.py
amrithasp02: I tried implementing the approach you suggested, and after comparing the "treetop_unique_ID" values before & after the change, the order/mapping in the final GeoDataFrame seems to be a bit different from what i had earlier. I'm trying to figure out which approach is going wrong, so it'll probably be good to discuss this live tomorrow.

tree_detection_framework/detection/region_detections.py
```diff
@@ -621,7 +621,14 @@ def merge(
             gdf[region_ID_key] = ID
 
         # Concatenate the geodataframes together
-        concatenated_geodataframes = pd.concat(detection_geodataframes)
+        concatenated_geodataframes = pd.concat(
+            detection_geodataframes, ignore_index=True
+        )
+
+        # Add a globally unique 5-digit ID column
+        concatenated_geodataframes["unique_ID"] = concatenated_geodataframes.index.map(
```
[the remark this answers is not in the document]
russelldj: Yeah, the auto-incrementing aspect seems really hard to deal with. That's why I was thinking random could be better, but that's also weird because the ID then has absolutely no meaning. An attribute of the detector that increments would be yet another approach.

PR #125, opened 2025-06-17
Move pixel-to-geospatial conversion from `RegionDetections` to `Detector`

### LINKED ISSUE #115: Move pixel-to-geospatial logic from `RegionDetections` to `Detector`
Currently, the `RegionDetections` object handles converting pixel space coordinates into geospatial ones. However, this has led to some pretty complex logic in the `__init__` [method](https://github.com/open-forest-observatory/tree-detection-framework/blob/67bcee63dd68f1f1bf70881b1a2334659f7bad22/tree_detection_framework/detection/region_detections.py#L170). We've never had a reason to access this pixel information after the RD is initially created, so we've discussed moving this logic into the `Detector` class. This could either be a method or a stand-alone function. It should likely be used somewhere around [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/67bcee63dd68f1f1bf70881b1a2334659f7bad22/tree_detection_framework/detection/detector.py#L126) to take the predictions in pixel space and convert them to geospatial, prior to creating the RD.

russelldj:
This PR closes #115 , which provides the context.

### DIFF
--- a/tree_detection_framework/detection/detector.py
+++ b/tree_detection_framework/detection/detector.py
@@ -7,11 +7,13 @@ import geopandas as gpd
 import lightning
 import numpy as np
 import pyproj
+import rasterio
 import shapely
 import torch
 from lightning.pytorch.callbacks import ModelCheckpoint
 from lightning.pytorch.loggers import TensorBoardLogger
 from scipy.ndimage import maximum_filter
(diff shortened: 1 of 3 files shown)


amrithasp02 (approved):
Looks great!

PR #128, opened 2025-06-23
Fix failure of `GeometricTreeCrownDetector` when tiles have no treetops

### LINKED ISSUE #127: `GeometricTreeCrownDetector` fails if tile has no tree tops
It looks like if the tile has no tree tops the `attribute` variable [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/64edb17bbaa89dda82774788e6ecca1198f2ef59/tree_detection_framework/detection/detector.py#L830) is just an empty dict. So you can't get expected attributes from it. One option would be to replace the indexing operator with `dict.get(...)` with an appropriate default value. @amrithasp02 I didn't get a chance to dig into this in any depth, do you have any thoughts on the best way to fix this?

amrithasp02:
This PR fixes #127


tree_detection_framework/detection/detector.py
```diff
@@ -825,6 +825,11 @@ def predict_batch(self, batch):
             # Set NaN values to zero
             image = np.nan_to_num(image)
 
+            # If attribute is empty, it means that the tile does not have any treetops
+            # detected. Such cases get ignored.
+            if not attribute:
```
russelldj, line 830: This is a lot simpler than adding dummy data to `batch_detections` and `batch_detections_data`. I imagine you thought about this, but I just wanted to confirm that you don't see any issues with variable-length batches of predictions? I would assume there's no assumptions on full batches anyway, because the last batch is going to be short in many cases. But now we'll just have to make sure we don't assume the input batch and output predictions will be the same length.

russelldj (approved):
Looks good, this fixed my issue on my data.

tree_detection_framework/detection/detector.py
russelldj: Great, thanks for doing that.

PR #130, opened 2025-07-01
Compute nonoverlapping core regions of a set of polygons and use for intelligent suppression of boundary detections

### LINKED ISSUE #176: Geometric tree detection is missing points and crowns at some edges
In some cases of two-stage geometric detection, there are missing detections at edge, in a straight line. Notably, there is a hard edge to the crowns, suggesting there may be tiles which are fully missing. This should be looked into more. An example from @youngdjn is below, showing this issue at the bottom.

Image

russelldj:
The goal of this is to make a set of potentially-overlapping polygons nonoverlapping. The definition for how to split is as follows: for any given point covered my multiple polygons, assign it to the polygon for which it is furthest interior from the nearest edge. The code so far only handles this for the pairwise case using the following steps:
- Identify the overlapping region
- Subtract this region from both input polygons to obtain the unambiguous regions
- Add additional points to the boundary using `segmentize`
- Create the veronoi tessellation to determine the region closest to each point
- Dissolve these regions based on which polygon boundary the point was from

I haven't rerun this recently, but I think it was having numerical issues.


tree_detection_framework/utils/geometric.py
```diff
@@ -142,3 +146,181 @@ def ellipse_mask(
...
+    voronoi_gdf.geometry = voronoi_gdf.make_valid()
+    merged = voronoi_gdf.dissolve("IDs")
+    merged["IDs"] = merged.index
+    if vis:
+        ax = merged.plot("IDs")
+        ax.set_title("Voronoi tessellation colored by original polygon")
+        plt.show()
+
+    poly1_merged = merged.iloc[0, :].geometry
+    poly2_merged = merged.iloc[1, :].geometry
```
amrithasp02: Is this just another way to do `iloc[0]` and `iloc[1]`?

tree_detection_framework/utils/geometric.py
```diff
@@ -142,3 +146,181 @@ def ellipse_mask(
...
+def make_polygon_set_nonoverlapping(
+    polygons: List[shapely.geometry.polygon.Polygon], epsilon: float = 1e-6, vis=False
+) -> List[shapely.geometry.polygon.Polygon]:
+    """
+    Take a set of polygons and return an updated set representing the "core" nonoverlapping regions
+    such that each core region represents the space which is farthest interior to that polygon,
+    compared to all other polygons.
+
+    Args:
+        polygons (List[shapely.geometry.polygon.Polygon]): _description_
```
amrithasp02: `_description_` placeholder to be replaced

tree_detection_framework/utils/geometric.py
```diff
@@ -142,3 +146,181 @@ def ellipse_mask(
...
+
+    for i, first_poly in enumerate(polygons):
+        for j, second_poly in enumerate(polygons[:i]):
+            first_poly_nonoverlapping, second_poly_nonoverlapping = (
+                split_overlapping_region(
+                    first_poly, second_poly, epsilon=epsilon, vis=vis
+                )
+            )
+            nonoverlapping_regions[i].append(first_poly_nonoverlapping)
+            nonoverlapping_regions[j].append(second_poly_nonoverlapping)
```
amrithasp02, line 306: Just wanted to confirm - if a polygon is not overlapping with anything, it still gets added to the final list, correct?

tree_detection_framework/utils/geometric.py
```diff
@@ -142,3 +146,181 @@ def ellipse_mask(
...
+
+    output_polygons = []
+
+    # Iterate over the keys in order to ensure output order matches input order
+    for i in sorted(nonoverlapping_regions.keys()):
+        all_polys = nonoverlapping_regions[i]
+        intersection = shapely.intersection_all(all_polys)
+
+        if isinstance(intersection, shapely.geometry.GeometryCollection):
+            poly_goems = [
```
amrithasp02: typo: `poly_goems`

tree_detection_framework/utils/geometric.py
```diff
@@ -142,3 +146,181 @@ def ellipse_mask(
...
+        NotImplementedError: If there are multiple polygons in an intersecting region.
+
+    Returns:
+        List[shapely.geometry.polygon.Polygon]: The core regions, ordered the same way as the input
+    """
+    nonoverlapping_regions = defaultdict(list)
+
+    for i, first_poly in enumerate(polygons):
+        for j, second_poly in enumerate(polygons[:i]):
+            first_poly_nonoverlapping, second_poly_nonoverlapping = (
```
amrithasp02, line 300: I understand that this method processes overlaps in polygon **pairs**. I am just curious how it handles an overlap with three or more polygons. It's probably correctly handled, I just didn't fully follow the logic.

amrithasp02 (approved):
This is really cool overall! Took me a while to understand but I think I followed the core logic and reason for doing this. I just have a few clarifying questions.

tree_detection_framework/utils/geometric.py
russelldj: Yeah, this is a bit weird. The overlap is computed exhaustively between every pair of polygons. Fortunately this is relatively cheap to compute in non-overlapping cases because it first checks whether there's any intersection. Then, for each original input polygon, you find the intersection of all non-overlapping regions, computed across all the pairs with other polygons. This should guarantee that there's no overlap with any other polygon in the final representation.

tree_detection_framework/utils/geometric.py
russelldj: Yes, in cases where a pair does not overlap, both input geometries will be returned unchanged.

tree_detection_framework/utils/geometric.py
russelldj: Yup, it was redundant. Changed.

PR #139, opened 2025-07-09
Fix zero-length dataloader 

amrithasp02:
This involved overriding the `__init__` method of `GridGeoSampler` so that it can also have one single tile covering the entire raster.

### DIFF
--- a/tree_detection_framework/preprocessing/preprocessing.py
+++ b/tree_detection_framework/preprocessing/preprocessing.py
@@ -11,8 +11,9 @@ import rasterio
 import shapely
 import torch
 from torch.utils.data import DataLoader
-from torchgeo.datasets import IntersectionDataset, stack_samples, unbind_samples
-from torchgeo.samplers import GridGeoSampler, Units
+from torchgeo.datasets import BoundingBox, GeoDataset, IntersectionDataset, stack_samples, unbind_samples
+from torchgeo.samplers import GeoSampler, GridGeoSampler, Units
+from torchgeo.samplers.utils import tile_to_chips, _to_tuple
 from torchvision.transforms import ToPILImage
 
 from tree_detection_framework.constants import ARRAY_TYPE, BOUNDARY_TYPE, PATH_TYPE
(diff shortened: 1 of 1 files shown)


tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -30,6 +37,60 @@
 )
 
 
+class UnboundedGridGeoSampler(GridGeoSampler):
```
russelldj, line 40: One option would be to make this sampler support both the original behavior and the new one, controlled by a keyword argument. I don't have a strong preference either way, but it might make it a little more explicit what the problem this class is addressing is.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -30,6 +37,60 @@
...
+    """Samples elements in a grid-like fashion.
+
+    This allows the full raster to be returned as a single tile.
```
russelldj, line 43: Nitpick: consider making this a bit more explicit about what the behavior of the base class is.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -234,10 +295,10 @@ def create_intersection_dataloader(
     raster_data = CustomRasterDataset(raster_data, **kwargs)
 
     # Create an intersection dataset that combines the datasets
```
russelldj, line 297: Nitpick: consider a comment about attributes being taken from the first dataset.

tree_detection_framework/preprocessing/preprocessing.py
```diff
@@ -30,6 +37,60 @@
...
+        GeoSampler.__init__(self, dataset, roi)
+        self.size = _to_tuple(size)
+        self.stride = _to_tuple(stride)
+
+        if units == Units.PIXELS:
+            self.size = (self.size[0] * self.res, self.size[1] * self.res)
+            self.stride = (self.stride[0] * self.res, self.stride[1] * self.res)
+
+        self.hits = []
+        for hit in self.index.intersection(tuple(self.roi), objects=True):
```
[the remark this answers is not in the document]
amrithasp02: Yep, added that

tree_detection_framework/preprocessing/preprocessing.py
amrithasp02: Added

PR #140, opened 2025-07-09
Fix invalid geometry during NMS

amrithasp02:
Error from @russelldj while doing NMS:

`shapely.errors.GEOSException: TopologyException: unable to assign free hole to a shell at 688231.5603850988 4376822.3865695409`

As expected, this was caused by invalid geometry in the predictions. Buffering the `geometry` column by zero almost always fixes such issues.

@russelldj the parameters were:
```
CHIP_SIZE = 1000
CHIP_STRIDE = 250
RESOLUTION = 0.2
```
I verified that it works after the changes.

### DIFF
--- a/tree_detection_framework/postprocessing/postprocessing.py
+++ b/tree_detection_framework/postprocessing/postprocessing.py
@@ -43,6 +43,9 @@ def single_region_NMS(
     # Extract the geodataframe for the detections
     detections_df = detections.get_data_frame()
 
+    # Buffer by zero to fix any issues with invalid geometries
+    detections_df["geometry"] = detections_df.geometry.buffer(0)
+
     # Determine which detections are high enough confidence to retain
     # Get rows that are both high confidence and not empty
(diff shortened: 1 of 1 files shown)


tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -43,6 +43,9 @@ def single_region_NMS(
     # Extract the geodataframe for the detections
     detections_df = detections.get_data_frame()
 
+    # Buffer by zero to fix any issues with invalid geometries
+    detections_df["geometry"] = detections_df.geometry.buffer(0)
```
russelldj, line 47: Do you know any considerations for choosing between `buffer(0)` and [make_valid](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.make_valid.html) for this operation?

russelldj (approved):
Thanks for looking into this

PR #141, opened 2025-07-15
Feature/fes/vis detections script

FranzEricSchneider:
Script to visualize a few TDF outputs (gpkg files).

Tested with
```
python3 tree_detection_framework/entrypoints/visualize_detections.py /ofo-share/scratch-eric/000936-01/00/ /ofo-share/scratch-eric/000936-01/rds-DF-ground-filtered/ /ofo-share/scratch-eric/000936-01/vis-rds-DF-ground-filtered/ --step 100 --n-images 14 --image-ext JPG
```

And it produced images like this:

### DIFF
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -22,6 +22,7 @@ def plot_detections(
     CRS: Optional[pyproj.CRS] = None,
     plt_ax: Optional[plt.axes] = None,
     plt_show: bool = True,
+    plt_points: bool = False,
     visualization_column: Optional[str] = None,
     bounds_color: Optional[Union[str, np.array, pd.Series]] = None,
     detection_kwargs: dict = {},
@@ -42,6 +43,8 @@ def plot_detections(
             A pyplot axes to plot on. If not provided, one will be created. Defaults to None.
         plt_show (bool, optional):
             Whether to plot the result or just return it. Defaults to True.
+        plt_points (bool, optional):
+            Whether to plot the centroid points within the boundaries. Defaults to False.
         visualization_column (Optional[str], optional):
             Which column to visualize from the detections dataframe. Defaults to None.
         bounds_color (Optional[Union[str, np.array, pd.Series]], optional):
@@ -65,7 +68,7 @@ def plot_detections(
 
     # If no axes are provided, create new ones
     if plt_ax is None:
-        _, plt_ax = plt.subplots()
+        _, plt_ax = plt.subplots(figsize=(10, 8))
 
     # Show the raster if provided
     if raster_file is not None:
@@ -87,6 +90,16 @@ def plot_detections(
     # Use the .boundary attribute to plot just the border. This works since it's a geoseries,
     # not a geodataframe
     bounds.boundary.plot(ax=plt_ax, color=bounds_color, **bounds_kwargs)
+
+    # Plot the centroids if requested
+    if plt_points is True:
+        for i, row in data_frame.iterrows():
+            centroid = row.geometry.centroid
+            x, y = centroid.x, centroid.y
+            plt_ax.plot(x, y, "ro", markersize=1)
+            plt_ax.text(
+                x + 3, y, f"{i:03}", color="red", fontsize=4, ha="left", va="top"
+            )
 
     # Show if requested
     if plt_show:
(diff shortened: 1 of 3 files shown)


tree_detection_framework/detection/region_detections.py
```diff
@@ -300,6 +313,7 @@ def plot(
         CRS: Optional[pyproj.CRS] = None,
         plt_ax: Optional[plt.axes] = None,
         plt_show: bool = True,
+        plt_points: bool = False,
```
russelldj, line 316: Could this be renamed to `show_centroid` as below? I think that's more descriptive.

tree_detection_framework/entrypoints/visualize_detections.py
```diff
@@ -0,0 +1,144 @@
...
+from rasterio.errors import NotGeoreferencedWarning
+from tqdm import tqdm
+
+from tree_detection_framework.detection.region_detections import RegionDetections
+
+# We know we don't care about these warnings from rasterio
+warnings.filterwarnings("ignore", category=NotGeoreferencedWarning)
+
+
+def visualize_detections(
```
russelldj, line 15: This is specific to image set detections (rather than a raster) so maybe the name should reflect that.

tree_detection_framework/entrypoints/visualize_detections.py
```diff
@@ -0,0 +1,144 @@
...
+    image_dir: Path,
+    detection_dir: Path,
+    out_dir: Path,
+    n_images: int,
```
russelldj, line 19: nitpick: it feels a bit redundant to have both `n_images` and `step`. Is there an advantage to having both, rather than just a step or sampling `n_images` evenly?

tree_detection_framework/entrypoints/visualize_detections.py
```diff
@@ -0,0 +1,144 @@
...
+        step: Step size between visualized images.
+        show_centroid: Whether to show centroid in visualization.
+        image_ext: Extension of image files (e.g., 'JPG', 'tif').
+    Returns: None. Saves files in the output directory.
+    """
+    # Get sorted lists of image and gpkg files
+    gpkg_paths = sorted(detection_dir.glob("*.gpkg"))
+
+    # Sample indices according to n_images and step
+    indices = [i for i in range(0, len(gpkg_paths), step)][:n_images]
```
russelldj, line 42: nitpick: might be simpler to use `np.arange` or similar to avoid the list comprehension.

tree_detection_framework/entrypoints/visualize_detections.py
```diff
@@ -0,0 +1,144 @@
...
+
+    for idx in tqdm(indices, f"Saving images to {out_dir}"):
+        gpkg_path = gpkg_paths[idx]
+        image_path = image_dir / f"{gpkg_path.stem}.{image_ext}"
+
+        detections = RegionDetections(
```
russelldj, line 48: Nice. At some point it might be worth having a standard way to load a `RegionDetections` from a file.

tree_detection_framework/entrypoints/visualize_detections.py
```diff
@@ -0,0 +1,144 @@
...
+    parser.add_argument(
+        "--image-ext",
+        type=str,
+        default="JPG",
+        help="Extension of image files (e.g., 'JPG', 'tif').",
+    )
+
+    args = parser.parse_args()
+
+    # Check that image_dir and detection_dir exist and are not empty
```
russelldj, line 112: nitpick: is there a reason this logic is in the parser rather than `visualize_detections`?

russelldj (changes requested):
Looks good!

tree_detection_framework/entrypoints/visualize_detections.py
FranzEricSchneider: I think of argument checking as the proper concern of parse_args. Like if an argument is supposed to be an int and you pass a string, parse_args will complain. If you are supposed to pass a directory of images and you don't, it's parse_arg's responsibility to complain.

This isn't a hard rule or anything, I just like it when the application logic doesn't have to do input quality checking.

tree_detection_framework/entrypoints/visualize_detections.py
FranzEricSchneider: Sure

tree_detection_framework/entrypoints/visualize_detections.py
russelldj: That makes sense, I wasn't thinking about wanting to visualize a small number of nearly sequential images.

PR #147, opened 2025-07-25
Feature/fes/postprocess filters

FranzEricSchneider:
This adds a post-processing filter which compares each set of image detections to a mask, and only keeps detections where a certain percentage of the polygon is "good" compared to the mask.

Testing:
1. A pytest test is created, checking that mask and detection overlaps of various degrees are filtered out correctly based on the threshold value
2. I ran ground mask filters on 4 images from a random mission, [visualized here](https://docs.google.com/presentation/d/1Jfl_z5veNrkcEeKV5yTtbLyJc3LssTDWNipmOJ4nOjc/edit?slide=id.g3717e1a5408_0_14#slide=id.g3717e1a5408_0_14). The masks are applied correctly, although whether we'd want to apply those masks is an open question.


tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+
+def remove_masked_detections(
+    region_detection_sets: Union[List[RegionDetectionsSet], List[PATH_TYPE]],
+    image_root: PATH_TYPE,
+    image_paths: List[PATH_TYPE],
+    valid_classes: List[int],
+    mask_root: PATH_TYPE,
+    mask_extension: str = ".png",
+    threshold: float = 0.4,
+) -> List[RegionDetectionsSet]:
```
amrithasp02, line 560: Should the return type also include `List[GeoDataFrame]`?

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+            def iterator(rds):
+                yield 0, gpd.read_file(rds)
+
+        # Iterate over the region detections, which is equivalent to iterating
+        # over the chips that detections were calculated in
+        for idx, gdf in iterator(rds):
+
+            # Get the mean value of each detection polygon, as a list of
+            # [{"mean": <value>}, ...]
+            stats = rasterstats.zonal_stats(
```
amrithasp02, line 663: This looks great. I was just wondering if this step could become very slow when there are large number of detections. Or does `zonal_stats` internally handle that efficiently?

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+    One example use case is that mask images could be rastered where the
+    certain pixels are marked as "ground" vs. "above ground". Then detections
+    that mostly overlap with the "ground" mask could be removed.
+
+    Args:
+        region_detection_sets (Union[List[RegionDetectionSet], List[PATH_TYPE]])
+            Each element is a RegionDetectionsSet derived from a specific drone image,
+            or a geospatial file containing the detections from a drone image.
+            Length is the number of raw drone images given to  the dataloader.
+        image_root (PATH_TYPE)
```
amrithasp02, line 572: Is there a reason to have this when you can take the common path from `image_paths` argument?

tree_detection_framework/postprocessing/postprocessing.py
FranzEricSchneider: I'm not quite sure what the requirements are, but it seems... okay? Using deepforest or detectree I think I would normally get low hundreds of detections per image. For example, here is a [pretty dense set of detections](https://docs.google.com/presentation/d/1Jfl_z5veNrkcEeKV5yTtbLyJc3LssTDWNipmOJ4nOjc/edit?slide=id.g37040a34c34_1_50#slide=id.g37040a34c34_1_50) that I think would be less than 300 overall? And the images are approximately 4000x5000 pixels.

Here's some timing data I took:
- 300 detections, (3000 x 5000) image size: 0.1187s
- 300 detections, (4000 x 5000) image size: 0.1255s
- 1,000 detections, (3000 x 5000) image size: 0.37s
- 1,000 detections, (4000 x 5000) image size: 0.37s
- 10,000 detections, (3000 x 5000) image size: 3.63s
- 10,000 detections, (4000 x 5000) image size: 3.53s

So I think image size isn't a driver, number of detections is. It's not lightning fast, but this seems like a workable speed to me.

pyproject.toml
```diff
@@ -17,6 +17,7 @@ scikit-image = "^0.24.0"
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^8.4.1"
+rasterstats = "^0.20.0"
```
russelldj, line 20: My understanding is the `dev.dependencies` group is intended for dependencies that are only required for development, like formatters, linters, test drivers, etc. Unless something has been reconfigured from when you (correctly) added `pytest` to the development group, `poetry add rasterstats` should work.

tests/test_postprocessing.py
```diff
@@ -0,0 +1,197 @@
...
+
+    for value, (minx, miny, maxx, maxy) in zip(cycle(valid_classes), corners):
+        mask[miny:maxy, minx:maxx] = value
+
+    # Convert to (H, W, 1) and save using PIL
+    mask_img = Image.fromarray(mask[:, :, None].astype(np.uint8).squeeze(), mode="L")
+    mask_img.save(path)
+
+
+class TestRemoveMaskedDetections:
```
russelldj, line 79: Nitpick: is there a reason for using a class here rather than individual functions? Given there's no class-level information. Completely understand if this is best practice for these types of tests.

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+    # Input checking
+    if len(image_paths) != len(region_detection_sets):
+        raise ValueError(
+            f"Mismatch between length of given image_paths ({len(image_paths)})"
+            f" and region detection sets ({len(region_detection_sets)})"
+        )
+
+    filtered_sets = []
+
+    # Iterate over [0] the path to a specific image and [2] the
```
russelldj, line 608: Nitpick: typo or I'm missing something in what `[0]` and `[2]` represent

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+        transform = Affine.translation(0, mask.shape[0]) * Affine.scale(1, -1)
+
+        # To save filtered region detections in a particular set
+        filtered_regions = []
+
+        # We should either have a RegionDetectionsSet class, or a file path
+        # to a dataframe containing the detections
+        if isinstance(rds, RegionDetectionsSet):
+
+            def iterator(rds):
```
russelldj, line 647: Clever fix

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+                for idx in range(len(rds.region_detections)):
+                    rd = rds.get_region_detections(idx)
+                    yield idx, rd.get_data_frame()
+
+        else:
+
+            def iterator(rds):
+                yield 0, gpd.read_file(rds)
```
russelldj, line 655: Why is the index 0 here and incrementing in the other case?

tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -542,3 +547,148 @@ def update_detections(detections: gpd.GeoDataFrame):
...
+
+        # Iterate over the region detections, which is equivalent to iterating
+        # over the chips that detections were calculated in
+        for idx, gdf in iterator(rds):
+
+            # Get the mean value of each detection polygon, as a list of
+            # [{"mean": <value>}, ...]
+            stats = rasterstats.zonal_stats(
+                gdf,
+                mask.astype(int)[::-1, ...],
```
russelldj, line 665: Can these conversions be precomputed before this for loop? Mostly for readability, but also because the `astype()` call will be creating a new array each time if the type needs to be changed.

tree_detection_framework/postprocessing/postprocessing.py
FranzEricSchneider: Whoops, I mean [0] and [1] as the two elements in the tuple → `impath` and `rds`

russelldj (approved):
Sorry for the delay, looks good to me!

russelldj:
Merged after discussing with @amrithasp02

PR #154, opened 2025-08-27
Add "treetop_unique_ID" column to watershed approach detections

amrithasp02:
This adds back `treetop_unique_ID` to the returned geometric tree crown detections. Also to be safe, I added logic to ensure the `treetop_unique_ID` column exists whenever IDs were provided, even if no crowns are detected or the GeoDataFrame is empty. In these cases, values are set to `NaN`

Sorry if it's a bit dense, but I hope this covers all edge cases now.


tree_detection_framework/detection/detector.py
```diff
@@ -798,8 +802,15 @@ def get_tree_crowns_watershed(
...
         # Create a gdf for the output
         crown_gdf = gpd.GeoDataFrame(
-            crowns, geometry="tree_crown", columns=["tree_crown", "treetop_height"]
+            crowns,
+            geometry="tree_crown",
+            columns=["tree_crown", "treetop_height", "treetop_unique_ID"],
         )
+
+        # Ensure the treetop_unique_ID column exists even if no treetop IDs were provided
+        if treetop_ids_provided and "treetop_unique_ID" not in crown_gdf.columns:
```
russelldj, line 811: I'm a bit confused about the logic here. Specifically, why is `treetop_unique_id` not guaranteed to be in the columsn? Does the `columns` argument on line 807 get overwritten when `crowns` actually has data? Or are the final `columns` the intersection of the `["tree_crown", "treetop_height", "treetop_unique_ID"]` and the keys in `crowns`, if crowns has data?

tree_detection_framework/detection/detector.py
```diff
@@ -798,8 +802,15 @@ def get_tree_crowns_watershed(
...
+            crown_gdf["treetop_unique_ID"] = pd.Series(dtype=str)
```
russelldj, line 812: Is this all empty strings?

tree_detection_framework/detection/detector.py
amrithasp02: Ah you're right. I think I was confused earlier and after checking again now it seems like this condition is unreachable - so I've removed it. 
```
        # Create a gdf for the output
        crown_gdf = gpd.GeoDataFrame(
            crowns,
            geometry="tree_crown",
            columns=["tree_crown", "treetop_height", "treetop_unique_ID"],
        )
```
For these lines, I also checked what happens when `crowns` does not have `treetop_unique_ID` values in the dict. It looks like that column gets filled with `NaN` values when `crown_gdf` gets created. Which seems fine to me because that would happen only when `GeometricDetector()` (not the sequential treetop and treecrown detection) is used and this class is not reporting the treetop IDs anyway ([link to code](https://github.com/open-forest-observatory/tree-detection-framework/blob/7e8a1a0b1588745bc97131275adfd7f801e37e45/tree_detection_framework/detection/detector.py#L1120))

russelldj (approved):

tree_detection_framework/detection/detector.py
amrithasp02: Yep that's right

PR #172, opened 2026-03-24
Updates to support polygon evaluation

amrithasp02:
The following updates were made:

1. `polygons_to_points()` is a new function that derives "representative points" for the polygon detections. There are 2 ways to produce these points: 1) Get centroid from each polygon 2) If a CHM is provided, assign the highest point as the "representative point" for that polygon (logic in `_chm_max_points()`). For either option, if a CHM is provided, this method will also create a new column called "heights" with height values from the CHM for the derived points. This step should be performed prior to `match_points()`, which can then be used without modification on the resulting points. 
2. Visualization method now includes an option to display the polygons underneath points, if present in the GeoDataFrame.
3. Added a new notebook to demonstrate this with SAM3 detections


tree_detection_framework/evaluation/evaluate.py
```diff
@@ -24,6 +27,84 @@
...
+    are populated when CHM sampling is performed.
+
+    Args:
+        detections: GeoDataFrame, RegionDetections, or RegionDetectionsSet with
+            polygon geometries to convert as points.
+        method: Method to use to derive the representative point from each polygon.
+            - "centroid": Use the centroid of each polygon. If `chm_path` is
+              provided, the CHM is sampled at each centroid to populate
+              `height_column`.
+            - "chm_max": Requires `chm_path`. The point is set to the
```
russelldj: I had a similar comment elsewhere, but I worry about trusting the CHM if any of it is nodata. Maybe instead default to the centroid if any pixels are nodata.

tree_detection_framework/evaluation/evaluate.py
```diff
@@ -24,6 +27,84 @@
...
+        - All other columns preserved unchanged.
+    """
+    # Convert input to GeoDataFrame if needed
+    if isinstance(detections, RegionDetectionsSet):
+        detections = detections.merge().get_data_frame()
+    elif isinstance(detections, RegionDetections):
+        detections = detections.get_data_frame()
+
+    # Check if height column already exists
+    if height_column in detections.columns:
```
russelldj: Nice

russelldj (commented):
Looks good, I just want to discuss the logic for polygons with partial CHM coverage.

tree_detection_framework/evaluation/evaluate.py
```diff
@@ -113,20 +194,110 @@ def _fill_in_heights(
...
+
+    with rasterio.open(chm_path) as src:
+        # Work in CHM CRS, then reproject points back to original CRS at the end
+        polys_in_chm_crs = gdf if gdf.crs == src.crs else gdf.to_crs(src.crs)
+        nodata = src.nodata
+
+        for i, (_, row) in enumerate(polys_in_chm_crs.iterrows()):
+            polygon = row.geometry
+
+            try:
```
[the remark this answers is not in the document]
amrithasp02: It's a safety net for `mask()` if it fails due to issues like invalid geometries, geometries fully outside the raster bounds, or any other raster related errors. It falls back to using the centroid.

sandbox/evaluation/evaluating_polygon_predictions.ipynb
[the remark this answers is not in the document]
amrithasp02: Thanks, changed it

tree_detection_framework/evaluation/evaluate.py
russelldj: Yes, that was my suggestion. Maybe we could instead use a threshold for the fraction of valid pixels? Are you seeing cases where there are individual invalid pixels scattered throughout the CHM? In that case I agree that my suggestion would definitely not make sense.

PR #174, opened 2026-03-26
Filter tree detections using heights from CHM

amrithasp02:
Added a new postprocessing method to remove detections that have CHM values lower than a height threshold. 
A detection is kept if either: 
- The max CHM pixel value within the polygon is ≥ min_height, or 
- There is no CHM data for that polygon because it is fully outside CHM bounds

A detection is removed if:
- The max CHM pixel value within the polygon is < min_height

Note: For polygons with partial CHM data, the missing pixels (w/o height values) are ignored and the decision is made from whatever valid data exists.

I also added a new visualization function to plot the CHM overlayed on the orthomosaic.

### DIFF
--- a/tree_detection_framework/postprocessing/postprocessing.py
+++ b/tree_detection_framework/postprocessing/postprocessing.py
@@ -823,10 +823,9 @@ def filter_by_chm(
             )
 
             # Extract max values, handle None -> np.nan
-            chm_heights = np.array([
-                stat["max"] if stat["max"] is not None else np.nan
-                for stat in stats
-            ])
+            chm_heights = np.array(
+                [stat["max"] if stat["max"] is not None else np.nan for stat in stats]
+            )
 
     # Keep detections if:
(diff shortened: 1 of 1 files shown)


tree_detection_framework/postprocessing/postprocessing.py
```diff
@@ -728,3 +730,111 @@ def iterator(rds):
...
+        # Reproject detections to match CHM CRS if needed
+        if detections_gdf.crs != chm_crs:
+            sampling_gdf = detections_gdf.to_crs(chm_crs)
+        else:
+            sampling_gdf = detections_gdf
+
+        chm_heights = []
+
+        # Find geometry type; note: this assumes all geometries are the same type
+        geom_type = sampling_gdf.geometry.iloc[0].geom_type
```
russelldj: You could check that there's in fact only one type with something like  `len(sampling_gdf.geometry.geom_type.unique())`

PR #178, opened 2026-04-07
Replace NMS with keeping central crowns in geometric detection

### LINKED ISSUE #175: Geometric tree crown detection may produce multiple crowns per tree top
The two-stage geometric detector works by first detecting tree top points (local maxima in the CHM) and then using watershed to segment tree crowns. Overlapping crowns are filtered at the boundaries of tiles using NMS ([here](https://github.com/open-forest-observatory/tree-detection-framework/blob/05306d5c162ea3a60fd38bc5d42c8089fa297d50/tree_detection_framework/entrypoints/detect_geometric_two_stage.py#L117)), with the intersection over self metric. This attempts to filter out duplicate crowns for the same tree top, but is not guaranteed to work properly, and multiple crowns for the same tree top can be introduced. An example is below. 

Image

This can result in confusing results for downstream tasks, such as per-crown species or live/dead prediction. A simple fix for this may be to take the largest crown per tree top (noted by `treetop_unique_ID`). It's possible this could fully replace NMS, or might need to be run after NMS.

russelldj:
This closes #175. It actually required only a small change to update the `distance` score to reflect the distance of the seed point, not the crown centroid. From there, NMS can be replaced by an approach that compares all crowns for a given tree top and retains the crown for which the tree top was farthest from the tile edge for the corresponding tile.

### DIFF
--- a/tree_detection_framework/utils/detection.py
+++ b/tree_detection_framework/utils/detection.py
@@ -34,7 +34,6 @@ def calculate_scores(
         raise ValueError(
             "Invalid confidence_feature provided. Choose from: `height`, `area`, `distance`, `all`"
         )
-
     if confidence_feature == "height":
         # Use height values as scores
         confidence_scores = tile_gdf["treetop_height"]
@@ -48,10 +47,6 @@ def calculate_scores(
         confidence_scores = tile_gdf[geometry_column].apply(lambda geom: geom.area)
 
     elif confidence_feature == "distance":
-        # Calculate the centroid of each tree crown
-        tile_gdf["centroid"] = tile_gdf[geometry_column].apply(
-            lambda geom: geom.centroid if not geom.is_empty else None
-        )
 
         # Calculate distances to the closest edge for each centroid
         def calculate_edge_distance(centroid):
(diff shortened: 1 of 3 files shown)


russelldj:
I tested this on `000450_000444` and the results look good. There are no duplicated `treetop_unique_ID`s. However, due to the recent change in handling boundary detections (#130), the tile bounds have shifted slightly and the `main` branch works as well. The command is below.

```
 python tree_detection_framework/entrypoints/detect_geometric_two_stage.py \
scratch/duplicate-crowns-debugging/000450_000444_chm-mesh.tif \
scratch/duplicate-crowns-debugging/tree-tops_original.gpkg \
--tree-crowns-save-path scratch/duplicate-crowns-debugging/tree-crowns_original.gpkg \
--chip-size 4000 \
--chip-stride 3800 \
--raster-blur-sigma 0.66 \
--resolution 0.12 \
--tree-top-detector-kwargs '{"b": 0.025, "c": 0.0625}' \
--edge-suppression-meters 10
```

image

amrithasp02 (approved):
Looks good to me!

tree_detection_framework/utils/detection.py
youngdjn: For clarity, I'd change `centroid` to `point` or similar.

youngdjn (approved):
Looks good! One small request to improve clarity.

PR #180, opened 2026-04-08
Retain crown for first (`00000`) treetop

### LINKED ISSUE #179: Tree detection approach with watershed does not produce crown for first detected tree top
The tree top with `unique_ID` `00000` does not have a corresponding tree crown, when running experiments on `000450_000444`. I seem to recall that @youngdjn mentioned a similar problem, but I cannot find the reference. I suspect this is related to watershed dropping the `0` ID, [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/572ea06fd816deb963a770bad6e7a02d7d226d4e/tree_detection_framework/detection/detector.py#L799).

Image

russelldj:
Closes #179 following @amrithasp02 's suggestion

I tested this on the example data with the following command on both `main` and this branch.
```
python tree_detection_framework/entrypoints/detect_geometric_two_stage.py \
data/emerald-point-chm/chm.tif scratch/missing-crown-polygons/tree-tops-main.gpkg \
--tree-crowns-save-path scratch/missing-crown-polygons/tree-crowns-main.gpkg \
--chip-size 1000 --chip-stride 900
```
This change correctly retained the crown for the treetop with ID `00000`. In both cases, three IDs ('00289', '01015', and '01413') did not have a crown. But I checked each of them and they correspond to tree tops that were short and on the "shoulder" of another tree, so it makes sense that no crown was identified.

### DIFF
--- a/tree_detection_framework/detection/detector.py
+++ b/tree_detection_framework/detection/detector.py
@@ -788,9 +788,11 @@ class GeometricTreeCrownDetector(Detector):
 
         # Rasterize treetops as markers
         markers = rasterize(
-            # Convert uid to int since it's a string with 0 padded values
-            [(pt, int(uid)) for pt, uid in zip(filtered_points, filtered_ids)],
+            # Convert uid to int since it's originally a string with 0 padded values
+            # Add 1 to the value because watershed expects the background to be represented by 0s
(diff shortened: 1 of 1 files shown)


amrithasp02 (approved):
Nice!

PR #181, opened 2026-04-10
Make nonoverlapping region computation work for single polygon

russelldj:
There was a bug in the nonoverlapping region computation that would produce a zero-length list if only one polygon was provided. This was because there were no pairwise comparisons, so the subsequent logic broke. This fix just returns the input one-length list unchanged. It doesn't really matter, but it similarly returns a zero-length list directly.

This issue can be seen where all detections will be suppressed if there's only one tile. The following command fails on `main` but works on this branch.
```
python tree_detection_framework/entrypoints/detect_geometric_two_stage.py data/emerald-point-chm/chm.tif \
  /tmp/tree-tops.gpkg --tree-crowns-save-path /tmp/crowns.gpkg --chip-size 4000 --chip-stride 3600 \
  --resolution 0.12 --raster-blur-sigma 0.333 --edge-suppression-meters 10
```

### DIFF
--- a/tree_detection_framework/utils/geometric.py
+++ b/tree_detection_framework/utils/geometric.py
@@ -148,12 +148,12 @@ def ellipse_mask(
     return mask.astype(bool)
 
 
-def ordered_voronoi(points):
+def ordered_voronoi(points, tolerance:float=1e-6):
     """
     The shapely version we are using does not order the voronoi polygons in the same way as the
     input points. This wrapper does that.
     """
-    voronoi = shapely.voronoi_polygons(points)
+    voronoi = shapely.voronoi_polygons(points, tolerance=tolerance)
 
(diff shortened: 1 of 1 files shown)


russelldj:
I added another small, mostly unrelated change of adding a `tolerance` parameter to the Voronoi call.. With the above command applied to the data at `/ofo-share/argo-data/argo-output/temp-dir/species-prediction-training-data-prep-7m4lr/0168_000450_000444/photogrammetry-products/000450_000444_chm-mesh.tif`, I was getting the following error

`ValueError: Multiple Voronoi polygons contain the point POINT (671667.6195114631 4371256.211323277)`

I wasn't able to figure out exactly what was going on, but adding this parameter fixes it.

amrithasp02 (approved):

PR #182, opened 2026-04-20
Inscribe circle in box predictions & eroding polygons

amrithasp02:
The PR makes 2 new additions to the polygon detections evaluation pipeline:

1. DeepForest produces box detections. To evaluate these accurately, a circle/ellipse will be inscribed within the box during the `polygons_to_points()` function call. The rest of the logic to determine points and heights remains the same.
2. Before finding the max CHM point in a polygon, the polygon can be eroded inwards by a specified distance to help exclude edge pixels just outside the tree crown.


tree_detection_framework/evaluation/evaluate.py
```diff
@@ -231,6 +258,13 @@ def _chm_max_points(
         for i, (_, row) in enumerate(polys_in_chm_crs.iterrows()):
             polygon = row.geometry
 
+            if erosion_distance > 0:
+                eroded = polygon.buffer(-erosion_distance)
```
russelldj: I doubt it matters much, but I suspect erosion is faster if done on the entire dataframe. You could implement the same logic for checking empties for an entire dataframe as well. 

That said, since you are doing it elementwise, you could do the erosion differently so you don't have to worry about it collapsing to nothing. One idea would be keep halving the erosion distance until it's non-zero. The other would be to take a fraction of the smallest dimension. For any reasonably shaped polygon, I think that would work.

tree_detection_framework/evaluation/evaluate.py
```diff
@@ -87,6 +92,12 @@ def polygons_to_points(
     # Store original polygon geometries in a new column before overwriting the active geometry with points
     result[crown_geometry_column] = result.geometry
 
+    # For box polygons (only DeepForest so far), replace geometry with the largest inscribed circle or ellipse before deriving points
+    first_geom = result.geometry.iloc[0]
+    # check if geometry is a box polygon (i.e. it is equal to its own bounding box)
+    if first_geom.equals(first_geom.envelope):
```
russelldj, line 98: Cool logic.

tree_detection_framework/evaluation/evaluate.py
amrithasp02: I moved the erosion operation outside the `for` loop and am now doing it at once for the `GeoDataFrame`. Does this look correct? Regarding polygons that collapse to nothing - I'm not sure if there is a possibility of this happening that frequently since the detection polygons would be of reasonable size. Only if the `erosion_distance` is very large (at that point it wouldn't be meaningful) this will be a problem. I do however like the idea to halve the erosion distance until it's non zero. If you still feel strongly that this is important to include, I'm happy to add it.

russelldj: Nice, that looks great. I'm good with leaving the current logic of a fixed erosion size.

russelldj (approved):

PR #184, opened 2026-04-30
Save RegionDetectionsSet as a multi-layer GPKG file

amrithasp02:
This PR adds a new method to save a `RegionDetectionsSet` object usually with the raw predictions. The previously used `save()` method was merging the predictions from all tiles, but we wanted a way to preserve the tile boundaries when saving to disk. 

To handle this, I added `save_tiled()` which creates a new layer for every tile's detected geometries and an additional bounds layer that stores a concatenation of bounds from all tiles in a `GeoDataFrame`. 

To read the multi-layer gpkg file from disk as a `RegionDetectionsSet` object, we can use the newly added classmethod `from_tiled_file()`.

Note: Tiles with zero detections are entirely skipped because `fiona` can't write them, but the tile's bounds are still recorded. Any tile without a detections layer is reconstructed as a `RegionDetections` with an empty `GeoDataFrame` and its bounds intact.


tree_detection_framework/detection/region_detections.py
```diff
@@ -669,6 +671,99 @@ def save(
...
+            bounds_records.append({"tile_id": i, "geometry": bounds_geom})
+
+        # Collect all bounds into a single layer
+        bounds_crs = (
+            CRS if CRS is not None else self.get_default_CRS(check_all_have_CRS=False)
+        )
+        bounds_gdf = gpd.GeoDataFrame(bounds_records, crs=bounds_crs)
+        bounds_gdf.to_file(save_path, layer="_bounds", driver="GPKG", mode="a")
+
+    @classmethod
```
russelldj, line 717: Clever to make this a classmethod.

tree_detection_framework/detection/region_detections.py
```diff
@@ -669,6 +671,99 @@ def save(
         # Save the data to the geofile
         concatenated_geodataframes.to_file(save_path)
 
+    def save_tiled(self, save_path: PATH_TYPE, CRS: Optional[pyproj.CRS] = None):
+        """Save to a GPKG file with one layer per tile plus a single shared bounds layer.
+
+        Layout:
+          - "detections_0", "detections_1", ... one detection layer per tile
```
russelldj, line 678: Tiny preference for zero padding these layer names.

tree_detection_framework/detection/region_detections.py
```diff
@@ -669,6 +671,99 @@ def save(
...
+        Args:
+            save_path (PATH_TYPE): Path to a .gpkg file written by save_tiled.
+
+        Returns:
+            RegionDetectionsSet: The reconstructed set with one RegionDetections per tile.
+        """
+
+        save_path = Path(save_path)
+        layers = fiona.listlayers(str(save_path))
+        detection_layers = sorted(
```
russelldj, line 730: Does this actually need to be sorted? I'm not seeing a reason why.

tree_detection_framework/detection/region_detections.py
russelldj: I see how this would require an extra step for converting to ints and you are already making sure you sort by the integer representation. So completely up to you.

tree_detection_framework/detection/region_detections.py
```diff
@@ -669,6 +671,99 @@ def save(
...
+            )
+            crs = bounds_gdf.crs
+
+            if i in detection_layer_map:
+                detections = gpd.read_file(save_path, layer=detection_layer_map[i])
+                crs = detections.crs
+            else:
+                # Empty tile doesn't have a "detections_{i}" layer
+                # Reconstruct it with an empty GeoDataFrame
+                detections = gpd.GeoDataFrame(geometry=gpd.GeoSeries([], crs=crs))
```
russelldj, line 755: Previously we could guarantee that all `RegionDetection` dataframes in a `Set` had the same columns, but this isn't the case anymore. I don't think this will cause an issue, even in concatenatation operations, but I just wanted to make sure you've thought about it.

russelldj (approved):
This looks great! I was initially expecting all the detections would be saved in the same layer (aggregated across regions) but I like this even better. Nice work testing it.

tree_detection_framework/detection/region_detections.py
amrithasp02: I also like zero padding generally. I thought about this and realized that to zero pad you would need to know the total tile count at save time to set the padding width (which may not be constant for all orthos) and also require some extra logic at load time. I'm not sure if it's worth the complexity.

tree_detection_framework/detection/region_detections.py
amrithasp02: I think `fiona.listlayers` doesn't guarantee that the layers are always returned in order so just added it to be safe

tree_detection_framework/detection/region_detections.py
amrithasp02: Noted, thanks for pointing that out

PR #185, opened 2026-05-11
Save all detections in the same layer - RegionDetectionsSet

amrithasp02:
Updated the `save_tiled` and `from_tiled_file` methods to support handling detections from all regions in the same layer. Previously it was creating a new layer for each region's detections.

### DIFF
--- a/tree_detection_framework/detection/region_detections.py
+++ b/tree_detection_framework/detection/region_detections.py
@@ -674,8 +674,8 @@ class RegionDetectionsSet:
         """Save to a GPKG file with a single detections layer (all tiles) and a bounds layer.
 
         Layout:
-          - "detections": one layer holding all detections with a "tile_id" column
-          - "bounds": one layer holding all tile bounds with a "tile_id" column
+          - "detections": one layer holding all detections with a "region_ID" column
+          - "bounds": one layer holding all tile bounds with a "region_ID" column
 
         This preserves tile structure (including empty tiles).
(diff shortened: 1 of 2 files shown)


tree_detection_framework/detection/region_detections.py
```diff
@@ -690,29 +689,40 @@ def save_tiled(self, save_path: PATH_TYPE, CRS: Optional[pyproj.CRS] = None):
...
             if len(detections) > 0:
-                mode = (
-                    "w" if first_write else "a"
-                )  # "a" appends a new named layer to the existing file
-                detections.to_file(
-                    save_path, layer=f"detections_{i}", driver="GPKG", mode=mode
-                )
-                first_write = False
+                detections["region_ID"] = i  # tag each row with its tile index
+                detection_frames.append(detections)
```
russelldj, line 702: You could potentially replace this section with the `merge` method [here](https://github.com/open-forest-observatory/tree-detection-framework/blob/1086a243034e3daeec85187d65a59084282adfb5/tree_detection_framework/detection/region_detections.py#L465), though it does give you an RDS and not the underlying geopackages.

russelldj (approved):

tree_detection_framework/detection/region_detections.py
amrithasp02: That's a good idea. However, it seems like `merge()` doesn't record the bounds the way we want it so it would still need a separate loop. `merge()` also adds a `unique_ID` column which I'm not sure is relevant here.

PR #186, opened 2026-05-11
Integrate TCDDetector

amrithasp02:
Added a new detector, `TCDDetector` from https://github.com/Restor-Foundation/tcd

Also updated `generate_predictions.py` and `Dockerfile` to include this new detector. The new image size is 10.8GB. I tested that the new detector works when using the docker image.


tree_detection_framework/detection/detector.py
```diff
@@ -1484,3 +1484,156 @@ def predict_batch(self, batch):
             )
 
         return all_geometries, all_data_dicts
+
+
+class TCDDetector(LightningDetector):
+    """Instance segmentation detector using the pretrained TCD (Restor Foundation) Mask R-CNN checkpoint.
+    This detector detecects 2 classes:
```
russelldj: typo: detecects

tree_detection_framework/detection/detector.py
```diff
@@ -1484,3 +1484,156 @@ def predict_batch(self, batch):
...
+        self.module = module
+        self.setup_predictor()
+
+    def setup_predictor(self):
+        """Build the Detectron2 model and load weights.
+
+        If cfg.MODEL.WEIGHTS is a HuggingFace repo ID (not a local path),
+        downloads model.pth from the Hub before loading.
+        """
+        import os
```
russelldj: Nitpick: imports should be at the top level unless they might lead to import errors

russelldj (commented):
I didn't get to a full review before the meeting, but I was wondering if there would be a way to share a lot more functionality with the existing `Detectree2Detector` since they rely on the same model? The only difference I saw so far is this approach has an associated class prediciton.

amrithasp02:
> I didn't get to a full review before the meeting, but I was wondering if there would be a way to share a lot more functionality with the existing `Detectree2Detector` since they rely on the same model? The only difference I saw so far is this approach has an associated class prediciton.

Good idea. I combined both `TCDDetector` and `Detectree2Detector` to be handled under the same `MaskRCNNDetector`. I think this makes it a lot cleaner & simpler.

I have updated and tested the notebooks & entrypoint after making this change.

tree_detection_framework/detection/detector.py
```diff
@@ -1469,16 +1467,11 @@ def predict_batch(self, batch):
...
-            # Get predicted classes
-            labels = instances.pred_classes.numpy()
-
             all_geometries.append(shapely_objects)
             all_data_dicts.append(
                 {
-                    "score": scores,
-                    "labels": labels,
+                    "score": instances.scores.numpy(),
+                    "labels": instances.pred_classes.numpy(),
```
russelldj, line 1474: How does this work if it's only a single-class prediction?

tree_detection_framework/detection/region_detections.py
```diff
@@ -749,8 +749,19 @@ def from_tiled_file(cls, save_path: PATH_TYPE) -> "RegionDetectionsSet":
                 region_id: group.drop(columns=["region_ID"]).reset_index(drop=True)
                 for region_id, group in all_detections.groupby("region_ID")
             }
+            # Build an empty GeoDataFrame with the same schema for tiles with no detections
+            schema_cols = all_detections.drop(columns=["region_ID"])
```
russelldj, line 753: Is this just an unrelated change?

tree_detection_framework/detection/region_detections.py
```diff
@@ -762,8 +773,8 @@ def from_tiled_file(cls, save_path: PATH_TYPE) -> "RegionDetectionsSet":
             if i in detections_by_tile:
                 tile_detections = detections_by_tile[i]
             else:
-                # tile was empty - reconstruct with an empty GeoDataFrame
-                tile_detections = gpd.GeoDataFrame(geometry=gpd.GeoSeries([], crs=crs))
+                # tile was empty - reconstruct with an empty GeoDataFrame matching the file schema
```
russelldj, line 776: Unrelated?

tree_detection_framework/entrypoints/generate_predictions.py
```diff
@@ -165,10 +174,13 @@ def generate_predictions(
...
-
-    if predictions_save_path:
-        # Save predictions to disk
-        outputs.save(predictions_save_path)
+        if predictions_save_path:
+            # Save predictions to disk
+            outputs.save(predictions_save_path)
+    else:
+        if predictions_save_path:
+            # Save without merging tiles
```
russelldj, line 182: Nice logic

russelldj (approved):
Looks good!

tree_detection_framework/detection/detector.py
amrithasp02: The `detectron2` config sets a `NUM_CLASSES` parameter which is 1 in case of `Detectree2` and is 2 for `TCD`. This is set when creating the detector module class. If it is a single class all labels will be 0.

tree_detection_framework/detection/region_detections.py
amrithasp02: Yes, sorry to include it in this PR. But it was a change I made to fix an issue I was having with reconstructing empty detection tiles when reading from the RDS file that was saved out

tree_detection_framework/detection/region_detections.py
amrithasp02: Yes, relevant to the issue commented above

PR #187, opened 2026-05-25
Add entrypoint scripts for TDF Argo workflow

amrithasp02:
Add detection and postprocessing entrypoint scripts to support TDF argo workflow.

This PR also includes one other unrelated change (refer comment below). Sorry for not making a separate PR.

### DIFF
--- a/tree_detection_framework/entrypoints/detect_trees.py
+++ b/tree_detection_framework/entrypoints/detect_trees.py
@@ -1,8 +1,13 @@
 import argparse
 import json
 from pathlib import Path
-from tree_detection_framework.entrypoints.detect_geometric_two_stage import detect_trees_two_stage
-from tree_detection_framework.entrypoints.generate_predictions import generate_predictions
+
+from tree_detection_framework.entrypoints.detect_geometric_two_stage import (
+    detect_trees_two_stage,
+)
+from tree_detection_framework.entrypoints.generate_predictions import (
+    generate_predictions,
+)
 
 
 def detect_trees(
--- a/tree_detection_framework/entrypoints/detect_trees.py
+++ b/tree_detection_framework/entrypoints/detect_trees.py
@@ -18,6 +18,17 @@ def detect_trees(
     detectree2_weights_path: str,
     sam3_checkpoint_path: str,
 ):
+    """
+    Detect trees using the specified detector and save raw detections.
+
+    Args:
+        detector: Detector name: geometric, deepforest, detectree2, sam2, sam3, or tcd.
+        detection_params: Detection parameters (chip_size, chip_stride, resolution, batch_size, etc.).
+        detector_dir: Directory for detector outputs. raw_detections.gpkg is written here.
+        preprocessed_local_files: Preprocessed local file paths (should have keys "ortho" and "chm").
+        detectree2_weights_path: Path to detectree2 pretrained weights (.pth file).
+        sam3_checkpoint_path: Path to SAM3 checkpoint file.
+    """
     ortho_path = preprocessed_local_files["ortho"]
     chm_path = preprocessed_local_files["chm"]
     output_path = detector_dir / "raw_detections.gpkg"
--- a/tree_detection_framework/entrypoints/detect_trees.py
+++ b/tree_detection_framework/entrypoints/detect_trees.py
@@ -18,8 +18,8 @@ def detect_trees(
     detectree2_weights_path: str,
     sam3_checkpoint_path: str,
 ):
-    ortho_path  = preprocessed_local_files["ortho"]
-    chm_path    = preprocessed_local_files["chm"]
+    ortho_path = preprocessed_local_files["ortho"]
+    chm_path = preprocessed_local_files["chm"]
     output_path = detector_dir / "raw_detections.gpkg"
 
     output_path.parent.mkdir(parents=True, exist_ok=True)
@@ -27,7 +27,7 @@ def detect_trees(
     if detector == "geometric":
         detect_trees_two_stage(
             CHM_file=Path(chm_path),
-            tree_tops_save_path=output_path, # add tree_crowns_save_path if you want to save tree crowns too
+            tree_tops_save_path=output_path,  # add tree_crowns_save_path if you want to save tree crowns too
             chip_size=int(detection_params.get("chip_size", 2000)),
             chip_stride=int(detection_params.get("chip_stride", 1900)),
             resolution=float(detection_params.get("resolution", 0.2)),
@@ -40,10 +40,18 @@ def detect_trees(
             chip_size=float(detection_params["chip_size"]),
             chip_stride=float(detection_params["chip_stride"]),
             tree_detection_model=detector,
-            resolution=float(detection_params["resolution"]) if detection_params.get("resolution") else None,
+            resolution=(
+                float(detection_params["resolution"])
+                if detection_params.get("resolution")
+                else None
+            ),
             predictions_save_path=str(output_path),
             run_nms=False,
-            batch_size=int(detection_params["batch_size"]) if detection_params.get("batch_size") else None,
+            batch_size=(
+                int(detection_params["batch_size"])
+                if detection_params.get("batch_size")
+                else None
+            ),
             detectree2_weights_path=detectree2_weights_path,
             sam3_checkpoint_path=sam3_checkpoint_path,
         )
--- a/tree_detection_framework/entrypoints/detect_trees.py
+++ b/tree_detection_framework/entrypoints/detect_trees.py
@@ -4,6 +4,9 @@ from pathlib import Path
 
 from tree_detection_framework.entrypoints.detect_geometric_two_stage import (
     detect_trees_two_stage,
+    CHIP_SIZE,
+    CHIP_STRIDE,
+    RESOLUTION,
 )
 from tree_detection_framework.entrypoints.generate_predictions import (
     generate_predictions,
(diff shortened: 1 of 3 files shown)


tree_detection_framework/utils/geometric.py
```diff
@@ -210,6 +210,16 @@ def split_overlapping_region(
     p1_min_inter = poly1.difference(intersection)
     p2_min_inter = poly2.difference(intersection)
 
+    # If either differenced region is not a proper Polygon with area, the Voronoi
```
amrithasp02, line 214: Changes in this file was made to fix issues encountered by some of the datasets while using [remove_detections_from_tile_overlap ](https://github.com/open-forest-observatory/tree-detection-framework/blob/a341a49af568bb20791b7c6556859a54363871dc/tree_detection_framework/postprocessing/postprocessing.py#L496) as a postprocessing step.

tree_detection_framework/entrypoints/detect_trees.py
```diff
@@ -0,0 +1,109 @@
...
+
+def detect_trees(
+    detector: str,
+    detection_params: dict,
+    detector_dir: Path,
+    preprocessed_local_files: dict,
+    detectree2_weights_path: str,
+    sam3_checkpoint_path: str,
+):
+    ortho_path = preprocessed_local_files["ortho"]
```
russelldj, line 35: Is there a reason to provide the ortho and CHM paths as a dict, rather than named arguments?

tree_detection_framework/entrypoints/detect_trees.py
```diff
@@ -0,0 +1,109 @@
...
+    chm_path = preprocessed_local_files["chm"]
+    output_path = detector_dir / "raw_detections.gpkg"
+
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+
+    if detector == "geometric":
+        detect_trees_two_stage(
+            CHM_file=Path(chm_path),
+            tree_tops_save_path=output_path,  # add tree_crowns_save_path if you want to save tree crowns too
+            chip_size=int(detection_params.get("chip_size", 2000)),
```
russelldj: Can these defaults be constants defined somewhere, perhaps in the detector [file](https://github.com/open-forest-observatory/tree-detection-framework/blob/main/tree_detection_framework/detection/detector.py)?

tree_detection_framework/entrypoints/detect_trees.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    print(f"[detect] Detections saved to {output_path}")
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Detect trees using the specified detector and save raw detections."
+    )
+    parser.add_argument(
```
russelldj, line 80: You can enforce `choices` here to make sure it's one of the arguments in the list.

tree_detection_framework/entrypoints/detect_trees.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def detect_trees(
+    detector: str,
+    detection_params: dict,
+    detector_dir: Path,
+    preprocessed_local_files: dict,
+    detectree2_weights_path: str,
+    sam3_checkpoint_path: str,
+):
```
russelldj, line 23: I know it's duplicative of the argparse help text but a docstring would be helpful.

tree_detection_framework/entrypoints/postprocess.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def postprocess(
+    postprocessing_id: str,
+    detector_dir: Path,
+    chm_path: str,
+    min_tree_height: float,
+    postproc_config_path: Path,
+):
+    raw_detections_path = detector_dir / "raw_detections.gpkg"
```
russelldj: I'd prefer to just provide both of these paths directly for more flexibility. Is there a reason to do it like this?

tree_detection_framework/entrypoints/postprocess.py
```diff
@@ -0,0 +1,125 @@
...
+from tree_detection_framework.evaluation.evaluate import polygons_to_points
+
+
+def postprocess(
+    postprocessing_id: str,
+    detector_dir: Path,
+    chm_path: str,
+    min_tree_height: float,
+    postproc_config_path: Path,
+):
```
russelldj, line 21: Docstring

tree_detection_framework/entrypoints/postprocess.py
```diff
@@ -0,0 +1,125 @@
...
+    raw_detections_path = detector_dir / "raw_detections.gpkg"
+    output_path = detector_dir / "detections.gpkg"
+
+    # Geometric: skip chain, just height filter
+    if not postprocessing_id:
+        detections = gpd.read_file(raw_detections_path)
+        print(f"[postprocess] Loaded {len(detections)} raw detections.")
+        n_before = len(detections)
```
russelldj: It looks like both geometric and CV have the same logic for height filtering and saving. Can you make this step happen once after the condition?

tree_detection_framework/entrypoints/postprocess.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if postprocessing_id not in full_config:
+            raise ValueError(
+                f"postprocessing_id '{postprocessing_id}' not found in {postproc_config_path}. "
+                f"Available keys: {list(full_config.keys())}"
+            )
+
+        steps = full_config[postprocessing_id]
+
+        for step in steps:
```
russelldj, line 67: I like this really procedural way of doing it.

tree_detection_framework/entrypoints/postprocess.py
```diff
@@ -0,0 +1,125 @@
...
+            fn_name = step["name"]
+            kwargs = step.get("args", {})
+
+            fn = getattr(postprocessors, fn_name, None)
+            if fn is None:
```
russelldj, line 72: Nice job managing the risk of using `getattr`

tree_detection_framework/entrypoints/postprocess.py
```diff
@@ -0,0 +1,125 @@
...
+            # filter_by_chm needs chm_path which is not in the yaml
+            if fn_name == "filter_by_chm":
+                kwargs["chm_path"] = chm_path
+
+            print(f"[postprocess] Applying {fn_name} with args {kwargs}")
+            result = fn(result, **kwargs)
+
+        # Convert polygons to points, sampling CHM max within each crown
+        print(f"[postprocess] Converting polygons to points via chm_max")
+        result = polygons_to_points(result, method="chm_max", chm_path=chm_path)
```
russelldj, line 87: Why not just specify this as another one of the postprocessing steps in `steps`?

tree_detection_framework/utils/geometric.py
russelldj: Good find. Do you happen to know if it was points or lines that was causing this issue?

tree_detection_framework/utils/geometric.py
```diff
@@ -239,8 +249,15 @@ def split_overlapping_region(
...
 
+    # If all points are collinear, voronoi_polygons fails with a GEOS LinearRing error
+    if all_verts.convex_hull.area == 0:
+        return (poly1, poly2)
+
     # Compute the voronoi tesselation with the polygons ordered consistently with the input verts
-    voronoi = ordered_voronoi(all_verts)
+    try:
+        voronoi = ordered_voronoi(all_verts)
+    except shapely.errors.GEOSException:  # have a fallback case in case of any failure
```
russelldj: I'd prefer this just fail. At minimum, can this be a warning?

tree_detection_framework/entrypoints/detect_trees.py
amrithasp02: The dict exists because `preprocessed_local_files` comes from the preprocess step's JSON output in the Argo workflow and it is convenient to pass it this way through multiple steps (detect-trees and postprocess). However, I do agree that at the function level it makes more sense for it to be named arguments.

tree_detection_framework/entrypoints/detect_trees.py
amrithasp02: I realized they are already defined as constants in [detect_geometric_two_stage.py](https://github.com/open-forest-observatory/tree-detection-framework/blob/a341a49af568bb20791b7c6556859a54363871dc/tree_detection_framework/entrypoints/detect_geometric_two_stage.py#L20). I'll import from there.

tree_detection_framework/entrypoints/detect_trees.py
amrithasp02: Nice, I didn't know we could do that

tree_detection_framework/entrypoints/postprocess.py
amrithasp02: I was not keeping them separate because within the Argo workflow all detector sub directories write the outputs with the same file name. But again, I can see how this is not ideal when using outside of Argo. So, I've added  `raw_detections_path` and `output_path` as additional optional args which are given preference. If they are not provided, then the code falls back to this behavior of deriving the path from `detector_dir`. Is that OK? [8660259](https://github.com/open-forest-observatory/tree-detection-framework/pull/187/commits/8660259d3dc38b2eb02c207e2acbb222e0efe4b5)

tree_detection_framework/entrypoints/postprocess.py
amrithasp02: Done, not sure how I missed that :)

tree_detection_framework/entrypoints/postprocess.py
amrithasp02: Since  `postprocessing_config.yaml` is a user defined list of postprocessors, there could be a risk of `polygons_to_points` not always being specified at the end of every postprocessing set. It is a step that has to happen for every detector (needed by evaluation) so I thought I'd specify it outside `steps` just to be safe.

tree_detection_framework/utils/geometric.py
amrithasp02: I'm pretty sure it was a line in my case

tree_detection_framework/entrypoints/postprocess.py
russelldj: That makes sense for now in the context of the Argo workflow. At some point we might want to make this entrypoint more generic though if people start using it for other tasks.

tree_detection_framework/entrypoints/postprocess.py
russelldj: Works for me!

tree_detection_framework/entrypoints/postprocess.py
russelldj: All good, that's what review is for.

russelldj (approved):

PR #190, opened 2026-07-20
Include 2 additional geometric detector parameters

amrithasp02:
It was required to specify `raster_blur_sigma` to do smoothing experiments for the gemetric detector. I also included `edge_suppression_meters` just in case it's needed for other future experiments.


russelldj (approved):
Looks good. Do you know why this required me to trigger the workflow run? It doesn't seem to be the case on the more recent PR and the docs make it sound like that should only be for PRs from external contributors.

amrithasp02:
> Looks good. Do you know why this required me to trigger the workflow run? It doesn't seem to be the case on the more recent PR and the docs make it sound like that should only be for PRs from external contributors.

Huh that's really confusing. I have no clue why it required that.

